PR #8, opened 2024-05-01
setup part

0xJepsen:
This is not complete, please don't merge. 

Next steps are to:
- [ ] Iron out our type definitions for Curves and Points (can we find a way to relate points to curves? Right now they are disjoint which makes me feel like the type theory for it could be better)
- [ ] We need a polynomial type, from #9


0xJepsen:
Closes for now

PR #14, opened 2024-05-02
redo the curve module

Autoparallel:
Now we have `CurvePoint` which must be a point on the curve via the assertion. Impled the Add trait instead of just an add method.


0xJepsen (approved):
lgtm

PR #20, opened 2024-05-02
feat: `PlutoField::primitive_root_of_unity()`

Autoparallel:
I provide a function here that gets the root of unity for our field (but can be generalized to any finite field).


src/field.rs
```diff
@@ -23,6 +23,14 @@ impl PlutoField {
   pub const ORDER_U32: u32 = PLUTO_FIELD_PRIME;
 
   pub fn new(value: u32) -> Self { Self { value } }
+
+  pub fn primitive_root_of_unity(n: u32) -> Self {
+    let p_minus_one = PLUTO_FIELD_PRIME - 1;
+    assert!(p_minus_one % n == 0, "n must divide p - 1");
+    let pow = PlutoField::from_canonical_u32(PLUTO_FIELD_PRIME - 1)
```
thor314, line 30: dead code

src/field.rs
```diff
@@ -23,6 +23,14 @@ impl PlutoField {
   pub const ORDER_U32: u32 = PLUTO_FIELD_PRIME;
 
   pub fn new(value: u32) -> Self { Self { value } }
+
+  pub fn primitive_root_of_unity(n: u32) -> Self {
```
thor314, line 27: can you comment why this works and what this does? i'm not familiar with what you've done here

PR #21, opened 2024-05-02
curve implementation updated

thor314:
- **C101 curve struct; CurveParams implemented for C101**
- **impl logic for curve c101**


0xJepsen (approved):
look lovely b

src/curve.rs
```diff
@@ -10,81 +12,97 @@ pub struct Curve<F: Field> {
...
-      two: <F as AbstractField>::from_canonical_u8(2),
-    }
-  }
+/// An Affine Coordinate Point on a Weierstrass elliptic curve
+#[derive(Clone, Debug, Copy)]
+pub struct AffinePoint<C: CurveParams> {
+  x:     C::FieldElement,
+  y:     C::FieldElement,
+  /// is the point the point at infinity
+  infty: bool,
```
Autoparallel, line 63: I think this should be an enum still. E.g., something like:
```suggestion
pub enum AffinePoint<C: CurveParams> {
  Point(C::FieldElement, C::FieldElement),
  Infinity,
```

Autoparallel (approved):

src/curve.rs
0xJepsen: This makes sense as a point in ONLY one of these two, which makes it feel more type theoretically correct. But it also might be a little more verbose. Could go either way.

src/curve.rs
```diff
@@ -10,81 +15,102 @@ pub struct Curve<F: Field> {
...
+  /// Integer field element type
+  type FieldElement: Field + Neg;
+  /// Order of this elliptic curve, i.e. number of elements in the scalar field.
+  const ORDER: u32;
+  /// Coefficient `a` in the Weierstrass equation of this elliptic curve.
+  const EQUATION_A: Self::FieldElement;
+  /// Coefficient `b` in the Weierstrass equation of this elliptic curve.
+  const EQUATION_B: Self::FieldElement;
+  /// Generator of this elliptic curve.
+  const GENERATOR: (Self::FieldElement, Self::FieldElement);
```
devloper, line 28: This will be the generator for g1, but when we get into field extensions, I guess we will also need a g2 generator? Perhaps those are represented elsewhere. 

Also curious how affine & projective coords live here.

devloper (approved):

devloper:
good improvement!

src/curve.rs
thor314: generator for the curve yes, so if we call this curve the first group, then yes, g1. I'm not sure whether there exists a pairing curve for C101.

PR #23, opened 2024-05-03
montgomery arithmetic

### LINKED ISSUE #16: feat: Modular Arithmetic Optimization
I think there are two modular arithmetic optimizations that would be good to add to the `field.rs` component. 

- [ ] #17
- [ ] #18

While they each have there distinct applications and performance improvements, since this is a learning resource, i believe we should at some point add both of them and well document each. 

There is a pretty good resource on both these algorithms with slides from risc0 [here](https://www.youtube.com/watch?v=hUl8ZB6hpUM&t=2307s)

### LINKED ISSUE #17: Montgomery modular arithmetic

### LINKED ISSUE #18: Barrett modular arithmetic

lonerapier:
closes #16 
closes #17 
closes #18 

It changes the following:

- [x] Adds montgomery arithmetic for field. I've kept the implementation closer to theory. Let me know if we should change it back to Plonky3's implementation.
- [x] Barret reduction

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -248,12 +248,12 @@ const fn to_monty(val: u32) -> u32 {
 /// reduction, and some bit shifts and masks since R is a power of 2, saving a costly division.
 ///
 /// # Examples
-/// ```
+/// ```ignore
 /// let N = 101;
 /// let a = to_monty(10);
 /// let b = to_monty(20);
 /// let c = from_monty(a * b);
-/// assert_eq(from_monty(c), 99);
+/// assert_eq!(from_monty(c), 99);
 /// ```
 fn from_monty(x: u32) -> u32 {
   let x = x as u64;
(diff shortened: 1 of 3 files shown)


src/field.rs
```diff
@@ -7,12 +7,16 @@ use std::fmt;
...
+};
 use serde::{Deserialize, Serialize};
 
 const PLUTO_FIELD_PRIME: u32 = 101;
-// const MONTY_BITS: u32 = 7;
-// const MONTY_MASK: u32 = (1 << MONTY_BITS) - 1;
-// const MONTY_MU: u32 = 80;
+const MONTY_BITS: u32 = 7;
+const MONTY_MASK: u32 = (1 << MONTY_BITS) - 1;
+const MONTY_MU: u32 = 19; // (-P^-1) mod 2^MONTY_BITS
```
0xJepsen, line 19: Nice work! I think functionally this PR is amazing. My main feedback is i think we should go hard on some comments since we plan to make this a learning resource. If you had to explain to someone what you where doing, who didn't have much prior experience how would you paint the intuition? What intuition made the optimization clique for you? Is there a small example we can include in the comments or another test? Personally i find if i can explain things well it really solidifies my understanding too :)

0xJepsen (changes requested):

src/field.rs
lonerapier: @0xJepsen added comments and examples for montgomery and barret reduction. Let me know if those are not clearly explained.

0xJepsen (approved):
this is awesome :) Thanks for adding those

lonerapier:
> Looks like a test isn't passing

fixed

PR #38, opened 2024-05-06
feat: home-baked `FiniteField` trait

### LINKED ISSUE #27: feat: Remove `p3_field` dependency
If we make our own field trait, we can have more control over everything. In particular we can we can fit the `primitive_roots` in the trait and then make it applicable in the polynomial for the FFT's.

Autoparallel:
This PR gets the following:
- [x] Restructures `field.rs` -> `field/mod.rs`.
- [x] Creates the `FiniteField` trait in `field/mod.rs`.
- [x] Renames `PlutoField` to `GF101` and implements `FiniteField` on it.

Closes #27


Autoparallel:
We could probably streamline this trait a lot as only the `generator()` method is really something that depends on the field in a way that needs outside computing.

PR #44, opened 2024-05-07
Extension field

lonerapier:
This PR closes issue #40 

It changes the following:

- adds 2 degree extension field for `PlutoField`. I've not made field extension generic over degree. let me know if that is required.

### DIFF
--- a/src/field/gf_101.rs
+++ b/src/field/gf_101.rs
@@ -83,6 +83,8 @@ impl FiniteField for GF101 {
   fn neg_one() -> Self { Self::new(Self::ORDER - 1) }
 
   fn generator() -> Self { Self::new(2) }
+
+  fn from_canonical_u32(n: u32) -> Self { Self::new(n) } // TODO: recheck
 }
 
 impl Add for GF101 {
--- a/src/field/gf_101.rs
+++ b/src/field/gf_101.rs
@@ -40,21 +40,6 @@ impl FiniteField for GF101 {
 
   const ORDER: Self::Storage = PLUTO_FIELD_PRIME;
 
-  #[inline]
-  fn pow(&self, power: Self::Storage) -> Self {
-    let mut current = *self;
-    let power = power as u64;
-    let mut product = Self::one();
-
-    for j in 0..(64 - power.leading_zeros()) as usize {
-      if (power >> j & 1) != 0 {
-        product *= current;
-      }
-      current = current * current;
-    }
-    product
-  }
-
   fn inverse(&self) -> Option<Self> {
     if self.value == 0 {
       return None;
@@ -84,7 +69,7 @@ impl FiniteField for GF101 {
 
   fn generator() -> Self { Self::new(2) }
 
-  fn from_canonical_u32(n: u32) -> Self { Self::new(n) } // TODO: recheck
+  fn from_canonical_u32(n: u32) -> Self { Self::new(n) }
 }
 
 impl Add for GF101 {
--- a/src/field/gf_101_2.rs
+++ b/src/field/gf_101_2.rs
@@ -6,13 +6,19 @@ use std::ops::{Div, DivAssign, MulAssign, Rem};
 
 use serde::{Deserialize, Serialize};
 
+use super::ExtensionField;
 use crate::field::FiniteField;
 
+/// Pluto curve with modulus 101 supports two degree extension field. This can be verified
+/// by finding out embedding degree of the curve, i.e. smallest k: r|q^k-1
 const EXT_DEGREE: usize = 2;
-const QUADRATIC_EXTENSION_FIELD_ORDER: u32 = 10201;
+const QUADRATIC_EXTENSION_FIELD_ORDER: u32 = 101 * 101;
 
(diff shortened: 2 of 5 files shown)


src/field/gf_101_2.rs
```diff
@@ -0,0 +1,335 @@
...
+};
+use std::ops::{Div, DivAssign, MulAssign, Rem};
+
+use serde::{Deserialize, Serialize};
+
+use super::ExtensionField;
+use crate::field::FiniteField;
+
+/// Pluto curve with modulus 101 supports two degree extension field. This can be verified
+/// by finding out embedding degree of the curve, i.e. smallest k: r|q^k-1
```
0xJepsen, line 13: When reading about the embedding degree I was trying to see this by hand but came across some questions. In our case r is our curve order which is 17 and p is field order which is 101. How does 17 | 101 ^2 - 1 which is 17 | 10200. 17 doesn't divide 10200. Also 10200 mod 101 is just 17 | 100. Do I have an error in my arithmetic? Are we working in a field here?

src/field/gf_101_2.rs
```diff
@@ -0,0 +1,335 @@
...
+const EXT_DEGREE: usize = 2;
+const QUADRATIC_EXTENSION_FIELD_ORDER: u32 = 101 * 101;
+
+#[derive(Clone, Default, Copy, Eq, PartialEq, Hash, Debug, Serialize, Deserialize)]
+
+/// Quadratic Extension field element represented as polynomial of degree 1 in form:
+/// a_0 + a_1*t where {a_0, a_1} \in \mathhbb{F}. Uses irreducible poly of the form:
```
0xJepsen, line 20: I think this is the right approach but curious how or if we should use this in our polynomial module. I know @Autoparallel you have been working on the FFTs and polynomial interface. Maybe you have some insight here?

src/field/gf_101_2.rs
```diff
@@ -0,0 +1,335 @@
...
+/// (X^2-K).
```
0xJepsen, line 21: should this be X^2 -K or X^2 + K?  X^2 + 2 is irreducible in F_{101} and should be used to generate our extension field.

src/field/gf_101_2.rs
```diff
@@ -0,0 +1,335 @@
...
+    let scalar =
+      (self.value[0].square() - Self::irreducible() * self.value[1].square()).inverse().unwrap();
+    res.value[0] = self.value[0] * scalar;
+    res.value[1] = -self.value[1] * scalar;
+    Some(res)
+  }
+
+  #[inline]
+  fn pow(&self, power: Self::Storage) -> Self {
+    // TODO: not ideal to duplicate across traits impls
```
0xJepsen, line 90: Maybe we can add a default impl in the trait what do you think?

src/field/gf_101.rs
```diff
@@ -83,6 +83,8 @@ impl FiniteField for GF101 {
   fn neg_one() -> Self { Self::new(Self::ORDER - 1) }
 
   fn generator() -> Self { Self::new(2) }
+
+  fn from_canonical_u32(n: u32) -> Self { Self::new(n) } // TODO: recheck
```
Autoparallel, line 87: This is a nit, but we don't really need this if we just use `GF101::new()` right?

src/field/gf_101_2.rs
Autoparallel: Let me get the polynomial PR up.

src/field/gf_101_2.rs
Autoparallel: I think both work as (correct me if I'm wrong) 2 is not a square in GF_{101}.

src/field/gf_101_2.rs
Autoparallel: Looking good!

Autoparallel (approved):
I'm good with this to go in as is :)

src/field/gf_101_2.rs
lonerapier: yeah, both of them are irreducible, added it in the field sage script as well

https://github.com/pluto/ronkathon/blob/9c0ba7055b0f80b0d157243ca48bf002ed823ca4/math/field.sage#L47-L51

src/field/gf_101.rs
lonerapier: but then that's struct specific function right? We need a trait specific function to create new values, for example: in extension fields, each subvalue in extension field element is base field element.

https://github.com/pluto/ronkathon/blob/9c0ba7055b0f80b0d157243ca48bf002ed823ca4/src/field/gf_101_2.rs#L69

Is there a better way in Rust to do this? Sorry, still in my learning path for Rust.

src/field/gf_101_2.rs
lonerapier: done

lonerapier:
@0xJepsen @Autoparallel changes done. let me know if any others are needed

src/field/gf_101_2.rs
0xJepsen: that's a classic, my bad.

PR #51, opened 2024-05-09
fix mul, generator and add hardcoded tests

lonerapier:
fixes mul and multiplicative generator of extension field

### DIFF
--- a/src/field/gf_101_2.rs
+++ b/src/field/gf_101_2.rs
@@ -67,10 +67,7 @@ impl<F: FiniteField> FiniteField for QuadraticPlutoField<F> {
   // f_2_primitive_element = F_2([2, 1])
   // assert f_2_primitive_element.multiplicative_order() == 101^2-1
   // ```
-  fn generator() -> Self {
-    // TODO: unsure if this is correct or not, research more
-    Self { value: [F::from_canonical_u32(2), F::from_canonical_u32(1)] }
-  }
+  fn generator() -> Self { Self { value: [F::from_canonical_u32(2), F::from_canonical_u32(1)] } }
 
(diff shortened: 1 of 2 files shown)


src/field/gf_101_2.rs
```diff
@@ -52,24 +50,26 @@ impl<F: FiniteField> From<F> for QuadraticPlutoField<F> {
 impl<F: FiniteField> FiniteField for QuadraticPlutoField<F> {
   type Storage = u32;
 
+  const NEG_ONE: Self = Self::new(F::NEG_ONE, F::ZERO);
+  // TODO: This is wrong
+  const ONE: Self = Self::new(F::ONE, F::ZERO);
```
0xJepsen, line 55: What should this be?

src/field/gf_101_2.rs
```diff
@@ -322,7 +340,7 @@ mod tests {
 
   #[test]
   fn test_generator_order() {
-    let generator = F2::generator();
+    let generator = F2::ONE;
     let mut x = generator;
     for _ in 1..F2::ORDER {
       x *= generator;
```
0xJepsen, line 346: This isn't correct right? We are just multiplying one together a bunch of times, this test would pass no matter what right?

0xJepsen (approved):

src/field/gf_101_2.rs
lonerapier: oops lol yes, how did i miss this. fixed it now. Now, we're multiplying 1 by generator, order-1 times, which is the multiplicative order of the group defined by field.

src/field/gf_101_2.rs
lonerapier: i think it's a clippy issue, it reorganised the comments somehow, not sure.

PR #52, opened 2024-05-09
chores: refactor + cleanup

Autoparallel:
Working to:
- reduce LOC
- reduce redundancies
- generalize some things
- fix some bugs
- make things easier to use
- Restructure crate
- Renames


0xJepsen (approved):
Very clean PR nice work, well documented too

devloper (commented):
scrub a dub dub, nice cleanup!

Autoparallel:
I could add the crate attribute to enforce docs or throw clippy warning. I think this would be good so we have to document everything including crate/module level and not just body level.

PR #58, opened 2024-05-10
feat: general extensions and documentation

### LINKED ISSUE #55: feat: general extensions via polynomial division
# Idea
We have `Ext2<F>` which allows us to build a degree 2 extension over a `FiniteField` given an irreducible polynomial. However, this is quite hard coded, and we could improve this interface along with the `ExtensionField` trait by using our `Polynomial`s which have implement `Rem`.

## Tasks
To get this done, let's break this up into a few small tasks.
- [ ] Replace the implementation of `Ext2` with a `Polynomial` as a field instead of `[F; 2]` where multiplication can be done modulo an irreducible `Polynomial`.
- [ ] **NEEDS MORE INFORMATION** Adjust traits as necessary and consider a general `Ext<N, F>` wrapper to house `N`th extensions over some field `F`.

--- 

At the moment, I don't know more specific tasks. I will take a look in the codebase and take more notes here as it unfolds.

Autoparallel:
This PR will close #55


src/curves/g1_curve.rs
```diff
@@ -1,23 +1,27 @@
+//! Elliptic curve `y^2=x^3+3`` over the prime field GF(101).
+
 use super::*;
 
+// TODO: An empty struct like this and G2Curve is a code smell. We should probably
```
0xJepsen, line 5: This was a design @thor314 had taken I believe from artworks? (correct me if I am wrong). I am okay thinking about a single struct too.

src/curves/mod.rs
```diff
@@ -1,83 +1,89 @@
...
-      AffinePoint::Infty => panic!("Cannot double point at infinity"),
+      AffinePoint::PointOnCurve(x, y) => (x, -y),
+      AffinePoint::Infinity => panic!("Cannot double point at infinity"),
     };
-    AffinePoint::new(x, C::FieldElement::ZERO - y)
+    AffinePoint::new(x, C::BaseField::ZERO - y)
   }
 }
+
+// TODO: This should likely use a `Self::ScalarField` instead of `u32`.
```
0xJepsen, line 80: not sure I see why? Scaler mult should work for any number (point can be doubled any amount of times), in the field or not. if we constrain the type here what does it give us? It seems like it would make the interface harder. I am down to turn it into the scalar field under the hood here, but that seems like an optimization

Autoparallel: My thought is that we should define a `ScalarField` for the curve anyway because we will eventually need access to this. It's a bit odd to multiply by a number that isn't in the scalar field of a curve, it's not really very intuitive to think about it that way to me at least since you should reduce `mod SCALAR_FIELD_ORDER` prior to computing it.

0xJepsen (approved):
Looks nice, a clean abstraction for the extension field!

src/curves/g1_curve.rs
Autoparallel: You could at least do a single struct like 
```rust
CurveGroup<F: FiniteField> {
    _phantom: F,
}
```
This way you can just use one struct and define it over whatever field you'd like.

I think there's probably something even nicer we could do to better utilize the `Curve` as we have as of now.

src/curves/mod.rs
0xJepsen: As far as i understand all elliptic curve points should be able to be multiplied by an arbitrarily large number. (under the hood this is just point doubling and point addition). No it's clear that it might be redundant to do so (especially if it is bigger than the scaler field order like you point out). But i think the interface should take any number and just abstractly mod the scaler order under the hood before point doubling. That just feels like a more flexible interface to me. But also i don't mind changing it since this is for learning.

PR #59, opened 2024-05-10
chore: cleanup `curve` module

Autoparallel:
Small refactor and removal of unused items in the `curve` module.

Edit: Going to do a bit more here!


0xJepsen:
hmmm yeah i don't know actually. There is maybe some benefit to thinking of each curve as a distinct object, but it is also the same curve over different fields. My autism is telling me to just implement like a group trait, but i know that is over kill. my curry howard context is getting stuck.
Thinking about the pairing functing it can only exist between two curve that are "pairing friendly" which technically means that the second curve is is the extension field with degree k where k is the smallest number such that the embedding degree r | p^k - 1. So maybe the pairing function should check that somehow? It will take in elements in g1, and g2 and then output something in some field (idk what field yet)
I guess my final answer is: It depends on whether we want to relate the two curve groups  as being pairing friendly or not.

0xJepsen (approved):
LGTM

PR #63, opened 2024-05-13
feat: (mostly) generic fields

### LINKED ISSUE #62: feat: Scalar field to the curve sub groups
The scaler field of the curve subgroups should be of order 17. So we should build a field of order 17 and add that associated type as the scalar field for the Curves.

Would be awesome to add scalar multiplication with this type too!

Autoparallel:
Makes fields much more generic. Cleans up file structure.

Closes #62


0xJepsen:
Nice did we want to add it curve groups and implement multiplication of points and the scalar field elements? or maybe hold off on that.

Autoparallel:
I can do that too

Autoparallel:
@0xJepsen @lonerapier check out how I went through the fields on this. They're upgraded quite a bit imo.

Things got a bit weird with the `GaloisField<N, P>`, but with a bit more work we could totally get all of this done completely generically and have all the constants we need built at compile time.

I also want to do the scalar field in this too. Lmk what you both think!

src/field/prime/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+  let mut i = 2;
+  while i * i <= n {
+    if n % i == 0 {
+      panic!("input is not a prime number");
+    }
+    i += 1;
+  }
+}
+
+const fn find_generator<const P: usize>() -> PrimeField<P> {
```
0xJepsen, line 88: some comments or a wiki link on what this algorithm is call or what the intuition here is would be sweet!

Autoparallel:
Added in type aliases now which is pretty helpful. Cut down quite a bit of excess.

@0xJepsen Do we want to implement the quadratic extension to `GF(17)` and then the `scalar_mult` stuff? I can do this, I just need to know if we really need it.

0xJepsen:
> Added in type aliases now which is pretty helpful. Cut down quite a bit of excess.
> 
> @0xJepsen Do we want to implement the quadratic extension to `GF(17)` and then the `scalar_mult` stuff? I can do this, I just need to know if we really need it.

I don't know enough yet to be able so say with lots of confidence. I think the scalar field of both our curve subgroups should be GF(17) (My intuition here is that the pairing checks are maps between the cyclic subgroups of the same size). So afaik we don't need the quadratic extension of GF(17).

Autoparallel:
> I don't know enough yet to be able so say with lots of confidence. I think the scalar field of both our curve subgroups should be GF(17) (My intuition here is that the pairing checks are maps between the cyclic subgroups of the same size). So afaik we don't need the quadratic extension of GF(17). Unless I misunderstood what you were asking here.

The pairing isn't a map on the scalar fields of the curve though, it's a map of $r$-torsion groups. As for the scalar field for the curve over the extension, it is (at least) GF(17^2), so if we need scalar multiplication on this curve I believe we should need that field.

0xJepsen:
> > I don't know enough yet to be able so say with lots of confidence. I think the scalar field of both our curve subgroups should be GF(17) (My intuition here is that the pairing checks are maps between the cyclic subgroups of the same size). So afaik we don't need the quadratic extension of GF(17). Unless I misunderstood what you were asking here.
> 
> The pairing isn't a map on the scalar fields of the curve though, it's a map of r-torsion groups. As for the scalar field for the curve over the extension, it is (at least) GF(17^2), so if we need scalar multiplication on this curve I believe we should need that field.

Right, r in our case is 17 right?, we are mapping between the 17-torsion subgroups?

Autoparallel:
> Right, r in our case is 17 right?, we are mapping between the 17-torsion subgroups?

I'm not sure why that is, is not, would need, or would not need to be the case. I believe we just need a $K$ such that $E(K)[r] \cong \mathbb{Z}_r \times \mathbb{Z}_r$. Choosing the specific value of $r$ is something I don't know the how or why. Do you?

Autoparallel:
@devloper added a README in `src/field/` that adds a bit more depth on what's going on here!

0xJepsen:
> is

Yeah this is both in the Lyn thesis under Tate pairings (I think we are doing a Tate pairing) and in Plonk by hand. In general you need $r$ and $q$ to be co-prime or else there is an attack that can solve the discrete log problem in linear time called an anomaly attack. (idk the details of it). since 17 and 101 are co-prime it is a valid choice. The question: Is there a smaller r that is co-prime to 101? is a good question that I haven't thought of yet. But I think r=17 is good and fine and also what they picked in plonk by hand. It also allows us to deal with a small $k = 2$ which is why we have a quadratic extension not a cubic (for k=3) or so on. If we change our $r$ to something else such that $k$ changes to satisfy $r | q^k -1$ then we might have change our extension field. This is my understanding.

0xJepsen (approved):
This looks good with the type aliases. I am curious what num_enum is and how you are using it.

PR #67, opened 2024-05-15
feat: Tate pairing

### LINKED ISSUE #56: Feat: Pairing Check
This is the last big feature we need to work on before we can verify our proofs and make some end-to-end tests.

The current knowns are:

- We have constructed pairing friendly curves from our extension fields. 

What we need to do now is understand what the output group $G_{T}$ is. I am currently under the impression that it will be some field.

A good reference for this is this [py-pairing](https://github.com/ethereum/py_pairing/blob/master/py_ecc/bn128/bn128_pairing.py) Vitalik made a while back. I think the next steps are going to be constructing the components for the juicy `miller_loop`

- [ ] `cast_point_to_field` (this field should be $G_{T}$
- [ ] curve `twist` (not sure if needs to be in curve 1 or curve 2)
- [ ] line function between two points. Should take in three points, two to form the line and a third to evaluate it at. 

after all of these we should have what we need to build the miller loop and do our pairing check!

Autoparallel:
Implements the Tate pairing for our curves. Adds in a test case which comes from Lynn's thesis. Also added the ability to draw random points from our base curve.

Closes #56 ?

### DIFF
--- a/src/curve/pairing.rs
+++ b/src/curve/pairing.rs
@@ -30,20 +30,20 @@ pub fn pairing<const R: usize>(
   todo!();
 }
 
-fn miller_loop<const R: usize>(
-  _p: AffinePoint<PlutoExtendedCurve>,
-  _q: AffinePoint<PlutoExtendedCurve>,
-) -> GaloisField<2, { PlutoPrime::Base as usize }> {
+fn miller_loop<C: EllipticCurve, const R: usize>(
+  _p: AffinePoint<C>,
+  _q: AffinePoint<C>,
+) -> C::BaseField {
   // Use the R to get a binary representation, then loop over the binary representation to do the
   // algorithm.
   todo!();
 }
 
-fn line_function<const R: usize>(
-  a: AffinePoint<PlutoExtendedCurve>,
-  b: AffinePoint<PlutoExtendedCurve>,
-  input: AffinePoint<PlutoExtendedCurve>,
-) -> GaloisField<2, { PlutoPrime::Base as usize }> {
+pub fn line_function<C: EllipticCurve, const R: usize>(
+  a: AffinePoint<C>,
+  b: AffinePoint<C>,
+  input: AffinePoint<C>,
+) -> C::BaseField {
   let (a_x, a_y) = match a {
     AffinePoint::Point(x, y) => (x, y),
     AffinePoint::Infinity => panic!("Cannot use point at infinity"),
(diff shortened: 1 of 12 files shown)


src/curve/pairing.rs
```diff
@@ -0,0 +1,76 @@
+//! Pairing operations for the Pluto curve.
+
+use super::*;
+
+// TODO: I believe we will definitely need to construct GF(17) outright, as we will need to perform
```
0xJepsen, line 5: I think this is right

src/curve/pairing.rs
```diff
@@ -0,0 +1,76 @@
...
+// arithmetic operations on it.
+
+// TODO: I think we also need to create something like a `struct Torsion<const R: usize>` to have a
+// representation of the r-torsion points on the curve. These may have to just be for our curve and
+// not generic as making them generic could be quite hard.
+
+/// Compute the Tate pairing of two points on the curve.
+pub fn pairing<const R: usize>(
+  p: AffinePoint<PlutoExtendedCurve>,
```
0xJepsen, line 14: should the input be from two different curve groups (asymmetric paring). If we look at the one in kzg we take a point in the F_q (the base field) and then a point in F_{q^2} (the extension field)

Autoparallel: The groups consist of points on the curve though :)

0xJepsen: Oh True! Should we restrict them at all? So that we know they are from the right groups.

src/curve/tests/fields.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+}
+
+impl ExtensionField<2, 59> for GaloisField<2, 59> {
+  const IRREDUCIBLE_POLYNOMIAL_COEFFICIENTS: [PrimeField<59>; 3] =
+    [PrimeField::<59>::ONE, PrimeField::<59>::ZERO, PrimeField::<59>::ONE];
+}
+
+/// Returns the multiplication of two [`Ext<2, GF101>`] elements by reducing result modulo
```
0xJepsen, line 56: This says GF101 but isn't it for `Ext<2, GF59>`?

src/curve/tests/fields.rs
```diff
@@ -0,0 +1,143 @@
...
+pub type TestField = PrimeField<59>;
+pub type TestExtension = GaloisField<2, 59>;
+
+impl FiniteField for TestExtension {
+  const ONE: Self = Self::new([TestField::ONE, TestField::ZERO]);
+  const ORDER: usize = TestField::ORDER * TestField::ORDER;
+  /// Retrieves a multiplicative generator for GF(101) inside of [`Ext<2, GF101>`].
+  /// This can be verified using sage script
+  /// ```sage
+  /// F = GF(101)
```
0xJepsen, line 12: Should also be for GF59 right?

src/curve/tests/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+
+mod fields;
+use fields::*;
+
+// Let's work through the example in Lynn's thesis so that we can be sure we compute the Tate
+// pairing correctly.
+
+// The curve is given by E: y^2 = x^3 + x over F_59.
+// The curve has 60 points and we will take r = 5 to get a group in the 5-torsion called G.
+// One generator for this group is P = (25,30) and we get:
```
0xJepsen, line 15: Nice this is good documentation

0xJepsen (approved):
remarkable, actually awesome work here. Very excited about this.

PR #68, opened 2024-05-16
Tiny rsa

### LINKED ISSUE #73: feat: Tiny RSA
a tiny implementation of RSA

0xJepsen:
This isn't necessarily in the scope but peaked my interest.

Will close #73 

One thing to be done on this would be to implement an algorithm for generating large primes and upgrade from tiny_rsa -> big rsa


Autoparallel:
Why not just use the `PrimeField` struct here instead of creating a whole new struct?

src/tiny_rsa/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    } else if power % 2 == 0 {
+      Self::new((self.pow(power / 2).value * self.pow(power / 2).value) % Self::ORDER)
+    } else {
+      Self::new((self.pow(power / 2).value * self.pow(power / 2).value * self.value) % Self::ORDER)
+    }
+  }
+}
+
+#[allow(dead_code)]
+/// Computes the modular inverse of e mod totient
```
Autoparallel, line 76: totient? Do you check that it is relatively prime

src/tiny_rsa/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+    let q = random_prime(p);
+    assert_ne!(p, q);
+    Self { value: value % (p * q) }
+  }
+}
+
+impl<const N: usize> FiniteField for RSAField<N> {
+  const ONE: Self = Self { value: 1 };
+  const ORDER: usize = N;
+  // placeholder, fields without prime order don't have a primitive element
```
Autoparallel, line 39: I'm confused by this a bit. What is an example of an RSA field that doesn't have a primitive element? If it isn't prime order, it can't be a finite field.

Autoparallel: In reading a bit more about RSA, you are doing modulo `pq` for two primes `p` and `q`. So I don't think you're working over a finite field in this case, just a cyclic group.

0xJepsen: > I'm confused by this a bit. What is an example of an RSA field that doesn't have a primitive element? If it isn't prime order, it can't be a finite field.

Yes that makes sense to me. I think i could try to finagle a little bit better of an abstraction for it. I was tinkering with the field traits we have, but they didn't feel quite right.

0xJepsen:
> Why not just use the `PrimeField` struct here instead of creating a whole new struct?

because the field is not prime!

src/tiny_rsa/mod.rs
Autoparallel: I don't know if you really want to work with this as a field since it is not a field

Autoparallel:
> > Why not just use the `PrimeField` struct here instead of creating a whole new struct?
> 
> because the field is not prime!

Then it is a ring not a field! Has to be of prime power order to be a field.

Autoparallel (approved):
it's everything i've ever dreamed of and more

PR #60, opened 2024-05-12
Feat/dsl

lonerapier:
This PR fixes/closes issue #53

Very rough right now, don't review.

PR #71, opened 2024-05-20
(WIP) feat: compiler

Autoparallel:
I made a small little DSL that will allow you to write Rust to generate the arithmetic expressions. My goal is to get this to yield circuits that can then be used along with our polynomial modules and etc.


Autoparallel:
@lonerapier Do you want to chat about this at all? It's brainworming me at the moment. Maybe we could cook together?

src/compiler/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+
+// Above seems done. Now we need to have a way to unravel a collection of expressions into a
+// circuit that may have the same inputs and outputs as the expressions. Inputs are going to be
+// the terminal variables found by fully unravelling expressions and they should be named. The
+// fully ravelled expressions are the outputs, and they can also be named
+use super::*;
+
+pub mod dsl;
+
+pub struct Wire {
```
lonerapier, line 28: how will wires have input and output? maybe a better name could be `Constraint`?

src/compiler/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+// the gates. This way gates cannot be improperly connected to other gates.
+pub struct Gate {
+  pub left:   Wire,
+  pub right:  Wire,
+  pub output: Wire,
+  pub op:     Operation,
+}
+
+pub struct Circuit {
+  pub gates: DiGraph<Gate, Wire>,
```
lonerapier, line 60: maybe instead of picturing a circuit as graph, a simple tuple of (Gate, Constraint) that extends to rows and column in execution trace?

lonerapier (commented):
absolutely love this direction. API based DSL will save a lot of errors that arise due to language parsing, and it's so easy to extend than separate language as in #60.

@Autoparallel would love to help in any way possible.

src/compiler/mod.rs
Autoparallel: oh yeah, that is better. I think I was trippin'

src/compiler/mod.rs
Autoparallel: I think I do want to trash the graph, this is a relic of my first pass. I will remove it for now. 

One thing I also want to do is impl a nice Display for our circuits, so that's on my mind as far as this goes.

src/compiler/mod.rs
lonerapier: Oh, got it. I got confused then. I remember previously it was names as `Connection` Maybe that made more sense? IDK, i'll let you cook.

PR #77, opened 2024-05-22
feat: ecdsa

### LINKED ISSUE #76: feat: tiny ECDSA
there are two types of signatures that need to be verified for x509 certificate verification: 
- ECDSA
- RSA
The tiny RSA here is a good example for RSA, but we should likely see if we can make a toy implementation of ECDSA to built a deep intuition on its security assumptions and implementation details.

0xJepsen:
closes #76


Autoparallel (approved):
hell yeah!! nice cookin', captain!

PR #80, opened 2024-05-25
feat: weil pairing

lonerapier:
It changes the following:

- add `sqrt` for `PlutoBaseFieldExtension`
- add weil pairing as test harness for tate pairing
- add more docs (because why not?)


Autoparallel (approved):
This is really awesome. I have no negative feedback whatsoever! Thank you for adding lots of docs and sage examples!

PR #83, opened 2024-06-01
feat: poseidon hash

### LINKED ISSUE #75: feat: Tiny Poseidon
a toy implementation of Poseidon.

lonerapier:
closes #75 

It changes the following:

- add Poseidon hash function 
- add sage script modified from [original](https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_params_poseidon.sage) constant generator script for testing purposes


src/hash/poseidon/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+/// * `sponge_state` describes current [`SpongeState`]. Initialised in [`SpongeState::Init`] state
+///   and, if sponge is squeezing, then it can't absorb new elements.
+/// * current new elements absorbed into hash state based on the sponge `rate`.
+pub struct SpongeConfig {
+  rate:          usize,
+  capacity:      usize,
+  sponge_state:  SpongeState,
+  absorb_index:  usize,
+  squeeze_index: usize,
+}
```
Autoparallel, line 170: One thing I would suggest here based on some checks you do later is use the [typestate pattern](https://zerotomastery.io/blog/rust-typestate-patterns/)

The reason why is that you can enforce that a sponge is actually a completely different *type* when it is in a different state. For instance, remove the enum and consider:

```rust
pub trait SpongeState {}

pub struct Init;
pub struct Absorbing;
pub struct Squeezing;
impl SpongeState for Init {}
impl SpongeState for Absorbing {}
impl SpongeState for Squeezing {}
```

Now, for instance, you define
```rust
pub struct PoseidonSponge<F: FiniteField, S: SpongeState> {...}
```

you definitely don't have to do this, but it would mean that you enforce that, at compile time, a sponge construction is being used properly as only specific states via a compile-time type can do specific actions.

impl<F: FiniteField> Posei

Autoparallel: Likewise, you can return a type that is now in a specific state too.

src/hash/poseidon/tests/constants.rs
```diff
@@ -0,0 +1,41 @@
...
+    vec![70, 22, 13, 47, 100, 75, 49, 65, 69, 77, 60, 86, 90, 97, 62, 45],
+    vec![67, 29, 58, 15, 95, 99, 6, 50, 97, 81, 19, 54, 57, 45, 83, 72],
+    vec![27, 97, 94, 9, 1, 52, 26, 19, 47, 22, 3, 4, 39, 15, 11, 46],
+    vec![74, 20, 89, 27, 94, 8, 81, 36, 70, 72, 76, 11, 15, 67, 19, 85],
+    vec![38, 73, 14, 8, 84, 53, 83, 45, 87, 19, 15, 41, 99, 96, 57, 76],
+    vec![9, 90, 1, 66, 88, 67, 14, 11, 18, 61, 30, 81, 36, 39, 4, 69],
+    vec![95, 72, 48, 70, 13, 87, 34, 82, 46, 56, 51, 62, 97, 20, 65, 24],
+  ];
+  (rc16, mds16)
+}
```
Autoparallel, line 41: Do you want to add any code to generate these constants? That would be pretty sweet to do in a later PR!

Autoparallel (approved):
This is looking super good! I left comments, but you don't have to take them or work on them, they were just suggestions. 

On a final note, I did get the SHA-256 merged, so you may want to rebase into that. I'm fine with calling the module `hash` or `hashes`, it makes no difference to me. I really like your README that was specific to Poseidon, I probably should have done that for SHA-256.

src/hash/poseidon/tests/constants.rs
lonerapier: yup, i've mentioned this in the readme as next steps. will tackle in a future PR for sure.

https://github.com/pluto/ronkathon/blob/eb6a1dec5662475cff25ff1a521824307fbe4086/src/hash/poseidon/README.md?plain=1#L121

lonerapier: I had modified original sage script to generate this rust file for testing purposes for now

0xJepsen (approved):
Over all i think this is really good. I think Colins feedback on the typestate pattern is good too.

src/hash/poseidon/mod.rs
Autoparallel: I like this pattern a lot. I find it really useful for most things that have some kind of "state" to them (hence typestate as a name lol). Anything that moves an error up to compile time is always preferred imo

src/hash/poseidon/tests/constants.rs
Autoparallel: Nice! Sorry i missed that in the README. Was too focused on the beautiful implementation :)

lonerapier:
@Autoparallel done with the changes, it looks really cool now, and most of the previous state transition errors are now detected at compile time which is so cool :)

thanks for this suggestion.

PR #88, opened 2024-06-05
feat: merkle tree and merkle proof

Autoparallel:
A very basic Merkle tree implementation that allows for Merkle proof of membership.


Autoparallel:
I did not write a README for this yet, but can later.

0xJepsen (approved):
I am down to merge this, should we add a readme here or in a later pr?

PR #89, opened 2024-06-07
constant polynomial arithmetic

lonerapier:
adds a degree bound as compile time constant generics


0xJepsen (approved):
Nice work on this! I will see about getting this merged this week. Excited about the work on binary fields too :)

PR #90, opened 2024-06-14
feat: binary field

lonerapier:
It changes the following:

- binary field
- binary field extensions using tower field arithmetic

I've kept the implementation closer to theory as described in Section 2.3 of [Binius](https://eprint.iacr.org/2023/1784.pdf). let me know if anything needs more work. 

TODO:
- [ ] figure out `primitive_element_generator` in extension fields
- [x] docs and tests
- [x] demonstrate power of binary fields by showing examples of embedding and small-by-large multiplication as described in the paper


src/field/binary/extension.rs
```diff
@@ -0,0 +1,305 @@
...
+use super::BinaryField;
+use crate::field::FiniteField;
+
+/// Binary extension field GF_{2^{2^K}} using binary towers arithmetic as explained in Section 2.3 of [DP23b](https://eprint.iacr.org/2023/1784.pdf)
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BinaryFieldExtension<const K: usize>
+where [(); 1 << K]: {
+  /// coefficients of field element
+  pub coefficients: [BinaryField; 1 << K],
+}
```
Autoparallel, line 16: This guy is a const expr fiend now!!

src/field/binary/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+};
+
+use crate::field::FiniteField;
+
+pub mod extension;
+#[cfg(test)] pub mod tests;
+
+/// binary field containing element `{0,1}`
+#[derive(Debug, Default, Hash, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BinaryField(u8);
```
Autoparallel, line 14: Perhaps a type alias instead with:
```suggestion
pub type BinaryField = PrimeField<2>;
```

Autoparallel: I believe that would collapse a lot of this module here.

src/field/binary/extension.rs
```diff
@@ -0,0 +1,305 @@
...
+  iter::{Product, Sum},
+  ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Rem, Sub, SubAssign},
+};
+
+use super::BinaryField;
+use crate::field::FiniteField;
+
+/// Binary extension field GF_{2^{2^K}} using binary towers arithmetic as explained in Section 2.3 of [DP23b](https://eprint.iacr.org/2023/1784.pdf)
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BinaryFieldExtension<const K: usize>
```
Autoparallel, line 12: I think if you take the type I suggested, you could basically work with this and think of it like a nested set of very specifically chosen `GaloisField<2,2>`. Almost like a `GaloiseField<GaloisField<...>>` type of object (I know this latter type would not compile, but hopefully it gets the point across).

src/field/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 use super::*;
 
+pub mod binary;
```
Autoparallel, line 5: perhaps tower here instead?

Autoparallel (commented):
This is looking good my friend!!!

Suggestions so far only. Hopefully nothing nitpicky :)

src/field/binary/mod.rs
lonerapier: that'll work flawlessly, but i believe it wouldn't allow us to showcase the main properties of binary field, i.e. addition being bitwise XOR and multiplication bitwise AND.

src/field/binary/extension.rs
lonerapier: everywhere I go, I see compile time constants now :)

lonerapier:
@0xJepsen @Autoparallel this is ready for review now, I have added binary fields using our ronk traits in tests. 

my reasoning being, I want to showcase the ease of arithmetic in binary fields and their efficient embedding and small by large multiplication properties. this obviously bloats the code a bit

can remove this entirely, and directly use our traits, if we feel that's the better approach.

0xJepsen (approved):
Very nice work on this sir

src/field/binary/mod.rs
Autoparallel: What if you use a trait impl override/default/specialization? There are some Rust unstable feats that may help here to let us specialize to the binary case.

I thought we had it in the project, but don't seem to anymore. Anyway, here is an RFC for it:
https://rust-lang.github.io/rfcs/1210-impl-specialization.html

There could be others.

src/field/binary_towers/extension.rs
```diff
@@ -0,0 +1,344 @@
...
+/// represented as vector of 2^K [`BinaryField`] components in multilinear basis,
+/// i.e. an element $b_v = \prod_{i=0}^{k-1}(v_i.X_i+(1-v_i))$ where $v_i$ is in GF(2).
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BinaryFieldExtension<const K: usize>
+where [(); 1 << K]: {
+  /// coefficient of field element
+  pub coefficients: [BinaryField; 1 << K],
+}
+
+impl<const K: usize> BinaryFieldExtension<K>
```
Autoparallel, line 26: Should I be thinking of `BinaryFieldExtension<K>` as the binary tower with height `K`?

Autoparallel (approved):
This looks good to me! I just had a few clarifying thoughts.

One on a previous comment, one in this review.

src/field/binary_towers/extension.rs
Autoparallel: Ah nice. Probably a nit, but what do you think of naming this `BinaryTower` so we can think of `GaloisField<K, 2>` as degree `K` extensions over `PrimeField<2>`?

Let it be known that it doesn't bother me either way, you should choose what you think is proper :)

src/field/binary/mod.rs
lonerapier: @Autoparallel i need some more time to check and iterate on this. so, i'll open a new PR for this. don't want to block AES implementations due to this change, so merging it.

Issue #100, opened 2024-06-23
DES encryption


lonerapier:
add DES encryption demonstrating possible attack vectors

PR #101, opened 2024-06-23
feat: DES encryption

lonerapier:
This PR fixes/closes issue #100 

It changes the following:

- adds DES symmetric encryption algorithm
- demonstrates various attacks possible:
  - known plaintext attack
  - weak key attack
  - bit complements


src/primitives/symmetric/encryption/des/mod.rs
```diff
@@ -0,0 +1,256 @@
+//! Contains implementation of DES encryption
+#[doc = include_str!("./README.md")]
```
Autoparallel, line 2: This is a really nice touch. We should add this everywhere

src/primitives/symmetric/encryption/des/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+  ///
+  /// let des = DES::new(secret_key);
+  ///
+  /// let message = rng.gen();
+  /// let encrypted = des.encrypt(&message);
+  /// let decrypted = des.decrypt(&encrypted);
+  /// ```
+  pub fn decrypt(&self, ciphertext: &Block) -> Block {
+    Self::des(ciphertext, self.subkeys.iter().rev())
+  }
```
Autoparallel, line 255: Perhaps these should be part of a trait? 

Would be really cool if we could have a proof that `decrypt(encrypt(msg)) = msg`, but, alas, we are in Rust land not  Lean land.

src/primitives/symmetric/mod.rs
Autoparallel, line 1: In this module we could add a trait for `Encryption` or something. Though if we do that, it's a little odd to have `symmetric::encryption` and perhaps we should reverse into `encryption::symmetric` and define an encryption trait in `encryption` module. Perhaps we ultimately have subtraits for symmetric/asymmetric? Thoughts?

Autoparallel (approved):
This PR is beautiful, seriously!  Amazing work. The README is so helpful, the code is well structured. I had to try hard to come up with suggestions, but I left a few for you if you want. By no means do you have to take them, I'm okay with merging this as-is.

src/primitives/symmetric/mod.rs
lonerapier: totally makes sense. i'll restructure according to this, and think on the encryption trait

src/primitives/symmetric/encryption/des/mod.rs
lonerapier: added to other modules as well

0xJepsen (approved):
This looks really good to me, i like the way this module is shaping up. Nice work!

PR #104, opened 2024-06-28
doc: fix doc attribute scope

eightfilms:
Applying the `doc` attribute without `!` just applies it to the next item, and does not actually include the `README.md`. For example:

Screenshot 2024-06-28 at 7 53 13 PM

This PR fixes that by using `#!` instead to apply it to the entire scope, thus actually including the contents in the README.

With this PR:

Screenshot 2024-06-28 at 7 53 35 PM

To test, run `cargo doc` within `ronkathon`.

edit: the latex rendering within the html seems janky, but this could be fixed in a followup PR.


0xJepsen (approved):
nice good work

eightfilms:
fixed errors - doctests were running the poseidon examples from the README (which we now ignore with `cfg_attr`)

PR #113, opened 2024-07-01
cleaner: reuse `.pow()` for inverse calculation in prime fields

supragya:
This PR fixes/closes issue: NA

It changes the following: Cleaner reuse of the `.pow()` function


0xJepsen (approved):
This is more idiomatic for sure

Issue #81, opened 2024-05-25
feat: Diffie-Hellman and elliptic curve DH


lonerapier:
Since we have ECDSA now, it would be fascinating to see a toy implementation of Diffie-Hellman and ECDH.

Resources:

- https://web.math.ucsb.edu/~kylehansen/Papers/Elliptic_Curve_Cryptography.pdf
- http://koclab.cs.ucsb.edu/teaching/ecc/project/2015Projects/Haakegaard+Lang.pdf
- https://www.gabriel.urdhr.fr/2021/10/19/diffie-hellman-intro/

jtriley-eth:
i can give this a go tomorrow, maybe move `src/ecdsa.rs` into `src/ecc/ecdsa.rs` then add the ecdh

lonerapier:
sir, if possible, do try tripartite DH using Tate pairings as well

jtriley-eth:
![ecdh](https://github.com/pluto/ronkathon/assets/41972979/d029054b-aa76-4490-8211-7846905307bd)

jtriley-eth:
gonna have to solidify my understanding of bilinear maps & G2+ a bit first, but will do 🫡

Autoparallel:
classic scary sounding thing for "take two things make new unique new thing"|

lonerapier:
"Key Exchange", simple and better.

@jtriley-eth just ping me, would love to jam with you and help anywhere possible. Pairings was an amazing topic that I enjoyed battling with a lot (still don't understand it completely lol)

devloper:
Pluto internal code has some magic ECDH that we depend on :).  Definitely recommend diving into and exploring this primitive.  It's key to how our web proofs works. 

We will be open-sourcing something substantial in ~2 weeks for that.

Autoparallel:
I want to push this up the list of TODOs. This is a really cool problem. @jtriley2p do you still have interest in this?

jtriley2p:
it was early july, a cool summer day, the aroma of rwandan espresso wafted through the air, its gentle notes of citrus and cherry accompanied an artisanal muesli, decorated with slices of orange and raspberries. i peered into my smart phone, the interface to the technological abyss, the screen was marked with "feat: Diffie-Hellman and elliptic curve DH", a familiar name. one which, in my amateur ventures into the shallows of the deep, dark ocean of cryptography, i thought quite approachable.

[...]

PR #115, opened 2024-07-01
Elliptic Curve Diffie Hellman

jtriley-eth:
> WIP

Adds ecdh, groups it with ecdsa in an ecc module, and adds a new README detailing its application to public key cryptography.

Adding support for one-round tripartite ecdh via tate pairings WIP.

This PR closes issue #81


jtriley-eth:
ok, so i'm thinking keep this one simple, just bipartite key exchange bc it's going to take a while to grasp the exact tripartite ecdh; seems like it's going to be more than *just* a tate pairing, which is already a bit fuzzy atm.

jtriley-eth:
moved to https://github.com/pluto/ronkathon/pull/125

PR #124, opened 2024-07-02
feat: AES decryption

### LINKED ISSUE #119: feat: AES decryption
Part of: #96 

AES encryption has been implemented in #102, but not decryption.

eightfilms:
Closes #119

Some key notes/changes:

- introduced `galois_multiplication` to be used for `mix_columns`: for encryption, we only needed to multiply column values once at worst to do a matrix multiplication for the polynomial a(x) = 3x^3 + x^2 + x + 2, so we could do it within `mix_columns()`, but since the inverse `inv_mix_columns`uses a more complex polynomial (the inverse of **a**,  a^-1(x) = 11x^3 + 13x^2 + 9x + 14), it would be cleaner to properly implement this fn.
- added README details for decryption.
- also changed `shift_rows` impl since I realized we could simply just use [`rotate_left`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.rotate_left) and [`rotate_right`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.rotate_right) found in Rust stdlib.


src/encryption/symmetric/aes/mod.rs
```diff
@@ -109,6 +136,31 @@ pub struct AES<const N: usize> {}
 #[derive(Debug, Default, Clone, Copy, PartialEq)]
 struct State([[u8; 4]; 4]);
 
+/// Multiplies a 8-bit number in the Galois field GF(2^8). This is done using "carry-less
+/// multiplication" or "bitwise multiplication", where the binary representation of each
+/// element is treated as a polynomial.
+///
+/// NOTE: this multiplication is not commutative - ie. A * B may not equal B * A.
+fn galois_multiplication(mut col: u8, mut multiplicant: usize) -> u8 {
```
lonerapier, line 144: just want to get others' opinions, is it possible to use $GF(2^{8})$ throughout, rather than have these specific implementations? 

I think this will make traversing the code a bit simpler, but might make other operations like `SubBytes`/`AddRoundKeys` more difficult.

lonerapier (approved):
Awesome work sir

src/encryption/symmetric/aes/mod.rs
eightfilms: Yeah that's a good point, in some parts of the code (such as this) we are indeed operating in $GF(2^8)$. I could see this change being easier to understand/read (since we already have `GaloisField` implemented), and that should really be the main goal.

0xJepsen: I think this would be a good idea. It would be nice to stand on the other primitives we have so we don't have to reinvent them every time

src/field/binary_towers/tests.rs
eightfilms, line 1: This big diff is just me moving the impls into a generalized `GaloisField<8, 2>`, now found in `gf_2_8.rs`

0xJepsen (approved):
Awesome work on this!

PR #125, opened 2024-07-05
mv ecdsa -> ecc, add ecdh

jtriley-eth:
Adds ecdh, groups it with ecdsa in an ecc module, and adds a new README detailing its application to public key cryptography.

This PR closes issue https://github.com/pluto/ronkathon/issues/81

WIP: tripartite ecdh + tate pairings


Autoparallel:
👀

PR #126, opened 2024-07-05
Migrate binary field from struct to enum.

jtriley-eth:
Refactors `BinaryField` from being a `usize` container to a sum type of `Zero` and `One`.

Also adjusts tests and binary field extensions.

### DIFF
--- a/src/field/binary_towers/extension.rs
+++ b/src/field/binary_towers/extension.rs
@@ -300,7 +300,7 @@ pub(super) fn multiply(a: &[BinaryField], b: &[BinaryField], k: usize) -> Vec<Bi
 
   // r1r2*X_{i-2}
   // X_{i-2}: a+b*x where a, b are vectors of length K/4, with a = 0, b = 1
-  let mut x_i_high = vec![BinaryField::new(0); 1 << (k - 1)];
+  let mut x_i_high = vec![BinaryField::Zero; 1 << (k - 1)];
   x_i_high[quarterlen] = BinaryField::ONE;
   let r1r2_high = multiply(&x_i_high, &r1r2, k - 1);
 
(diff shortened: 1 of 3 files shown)


supragya:
Curious if this is the best way for representation here. Would you create an enum for ternary field too with three elements inside?

Generally, goes for any F_p if p is known btw.

I guess for consistency's sake this is not the best possible thing.? But I may be wrong.

jtriley-eth:
it's tricky

enshrined numeric data types tend to emerge from the underlying processor, $2^8, 2^{16}, 2^{32}, 2^{64}, ...$, the `usize` wrapper makes sense from a hardware perspective, and would likely be more performant since it's corresponding to XOR and AND cpu instructions rather than pattern matching branches.

but from a type theoretic perspective, a sum type is the cleanest well-constrained representation of a binary field element. it uses the type checker to signal to the compiler the valid states of the type. the existing implementation enforces two states in its construction, but not its implementation. that is to say while you can't construct an invalid field element via `BinaryField::new`, it is possible to construct or mutate one through other functions as, by definition, its cardinality depends on `usize`.

a ternary field and, by extension, all prime fields, would ideally be defined in this way. using product types wouldn't work for prime fields bc the cardinality of a product type is definitionally not prime. however, obviously this doesn't scale in rust. recursive types are poorly handled and a prime field sum type with literally `115792089237316195423570985008687907853269984665640564039457584007908834671663` elements is ofc impossible.

so there's a balance to be struck. personally, i'd write a ternary field in the same way (sum type), but anything beyond single or low double digit primes is off the table imo. however, i do think the prime field type should be generalized in a way that accepts arbitrary internal field element types to generalize. ie `PrimeField<BinaryField>`, `PrimeField<U256>`, etcetc. ofc there's no way to really constrain the huge primes the same way, but i think where it's feasible we should constrain it to a sum type.

(or we could just do `BinaryField(bool)` lol)

0xJepsen (approved):

PR #135, opened 2024-07-24
Feat/groups

lonerapier:
It changes the following:
- `FiniteGroup` trait
- `MultiplicativePrimeGroup` struct
- `CurveGroup` trait
- refactor ECDSA to generic group

### DIFF
--- a/src/field/group.rs
+++ b/src/field/group.rs
@@ -7,10 +7,13 @@
 //! 4. Existence of inverse element for every element of the set: $a\oplus b=0$
 //! 5. Commutativity: Groups which satisfy an additional property: *commutativity* on the set of
 //!    elements are known as **Abelian groups**.
+//!
+//! [`FiniteGroup`] trait is implemented for all finite groups.
 use std::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 
 use super::{prime::find_primitive_element, FiniteField, PrimeField};
 
+#[const_trait]
 /// Group trait with finite number of elements
 pub trait FiniteGroup:
   std::fmt::Debug
@@ -44,7 +47,7 @@ pub trait FiniteGroup:
 
   /// Multiplication with the scalar element of the group, i.e. repeatedly applying group
   /// [`FiniteGroup::operation`] `b` number of times.
-  fn scalar_mul(&self, b: &Self::Scalar) -> Self;
+  fn scalar_mul(&self, scalar: &Self::Scalar) -> Self;
 }
 
 /// Additive group on integer $\mathbb{Z}_p$ modulo over prime `P`.
@@ -67,6 +70,7 @@ impl<const P: usize> FiniteGroup for AdditivePrimeGroup<P> {
 
   fn inverse(&self) -> Self { Self(Self::ORDER - self.0) }
 
+  /// naive scalar multiplication
   fn scalar_mul(&self, b: &Self::Scalar) -> Self {
     let mut res = *self;
     for _ in 0..b.value {
@@ -114,7 +118,7 @@ impl<const P: usize> MulAssign<PrimeField<P>> for AdditivePrimeGroup<P> {
   fn mul_assign(&mut self, rhs: PrimeField<P>) { *self = *self * rhs; }
 }
 
-/// [`FiniteGroup`] under multiplication implemented as integer, $Z/nZ$ modulo prime `r`.
+/// [`FiniteGroup`] under multiplication implemented as integer, $Z/rZ$ modulo prime `r`.
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
 pub struct MultiplicativePrimeGroup<const P: usize>(usize);
 
@@ -186,19 +190,30 @@ mod tests {
 
     // commutativity
     assert_eq!(one + ident, ident + one);
-    assert_eq!(one, ident);
+    // inverse
+    assert_eq!(one + one.inverse(), ident);
+    // associativity
+    assert_eq!(one + (ident + one), (one + ident) + one);
+    // scalar multiplication
+    assert_eq!(one * PrimeField::<17>::new(2), one + one);
   }
 
   #[test]
   fn mul_group_properties() {
-    type MulGroup = MultiplicativePrimeGroup<5>;
-    type ScalarField = PrimeField<5>;
+    type MulGroup = MultiplicativePrimeGroup<17>;
+    type ScalarField = PrimeField<17>;
 
-    let gen = MultiplicativePrimeGroup::<5>::GENERATOR;
+    let gen = MultiplicativePrimeGroup::<17>::GENERATOR;
 
     let ident = MulGroup::IDENTITY;
 
-    assert_eq!(gen + ident, gen);
+    // commutativity
+    assert_eq!(gen + ident, ident + gen);
+    // inverse
+    assert_eq!(gen + gen.inverse(), ident);
+    // associativity
+    assert_eq!(gen + (ident + gen), (gen + gen) + ident);
+    // scalar multiplication
     assert_eq!(gen * ScalarField::new(2), gen + gen);
   }
 }
--- a/src/field/group.rs
+++ b/src/field/group.rs
@@ -50,74 +50,6 @@ pub trait FiniteGroup:
   fn scalar_mul(&self, scalar: &Self::Scalar) -> Self;
 }
 
-/// Additive group on integer $\mathbb{Z}_p$ modulo over prime `P`.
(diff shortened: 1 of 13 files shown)


src/field/group.rs
```diff
@@ -0,0 +1,135 @@
...
+  + SubAssign
+  + Neg
+  + Mul<Self::Scalar, Output = Self>
+  + MulAssign<Self::Scalar>
+  + Clone
+  + Copy
+  + Default
+  + Eq {
+  /// Scalar defined in the group
+  type Scalar: FiniteField;
```
Autoparallel, line 33: This is confusing to me.

Types used in a group could realistically be anything. I think this is okay if it is a curve group, but would prefer to not typecast a general `FiniteGroup` to operate on `FiniteField`

For instance, consider the abstract (dihedral) group (presentation) `<r,f | r^n, f^2, rfrf>`. One could represent this group with two chars (or an enum with two variants) and define the operations via reduction on these rules.

src/field/group.rs
```diff
@@ -0,0 +1,135 @@
...
+  /// Identity element of group
+  const IDENTITY: Self;
+  /// total number of elements in group
+  const ORDER: usize;
+  /// primtive element of group
+  const GENERATOR: Self;
```
Autoparallel, line 39: Not every finite group has a generator!

Autoparallel (changes requested):
I left a few comments, nothing major.

Given those comments and the general structure, may I suggest the following refactor and generalization?

Consider creating an `algebra` module with:
- `group`
- `ring`
- `field`
As submodules. (You can leave `ring` as TODO).

Now, it is probably reasonable to then define the abstract notion of `group` and `field` and then use these as super traits for the more refined items such as `FiniteGroup`, `FiniteCyclicGroup`, `FiniteField`, etc.

src/field/group.rs
lonerapier: You're right. Changed it now to abstract type which can be defined by the implementation.

src/field/group.rs
lonerapier: Added a new trait `FiniteCyclicGroup` as derived trait of FiniteGroup that defines a generator.

lonerapier:
@Autoparallel Thanks for the review.

- refactored as `algebra` module with separate submodules for group, field, ring (in future).
- Tried adding `Group`, `Field` trait separately, but we need function overriding in derived traits because `operation` in `Group` and `FiniteGroup` can be done in different ways.
  - Is there any other way that we can define `Group` trait's operation and `FiniteGroup` trait's operation separately?

src/field/group.rs
Autoparallel: You da bomb

Autoparallel:
@lonerapier let me take a look.

One thing we could also consider is making the ring/field `+` op be a `AbelianGroup`. Food for thought. 

Give me a sec to go look through this.

src/algebra/README.md
```diff
@@ -0,0 +1,25 @@
+# Algebra
+
+[Groups](https://en.wikipedia.org/wiki/Group_(mathematics)) $G$ are algebraic structures which are set and has a binary operation $\oplus$ that combines two elements $a, b$ of the set to produce a third element $a\oplus b$ in the set.
+The operation is said to have following properties:
+1. Closure: $a\oplus b=c\in G$
+2. Associative: $(a\oplus b)\oplus c = a\oplus(b\oplus c)$
+3. Existence of Identity element: $a\oplus 0 = 0\oplus a = a$
+4. Existence of inverse element for every element of the set: $a\oplus b=0$
+5. Commutativity: Groups which satisfy an additional property: *commutativity* on the set of
+   elements are known as **Abelian groups**.
```
Autoparallel, line 10: This is a nitpick, so don't feel compelled to change this (i love your documentation always) but I have a disdain for the $\oplus$ as a group op. Why? 

$\oplus$ is used, usually, as a *direct sum* or *coproduct* so it is a bit weird. It also carries connotation of commutivity which is usually *not* guaranteed in groups! (most groups are not commutative)

Instead consider just using concatenation: $ab = c$, $(ab)c=a(bc)$, etc.

src/algebra/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+[Rings](https://en.wikipedia.org/wiki/Ring_(mathematics)) are algebraic structures with two binary operations $R(\oplus, \cdot)$ satisfying the following properties:
+1. Abelian group under $\oplus$.
+2. Monoid under $\cdot$.
+3. Distributive with respect to $\oplus$.
+
```
Autoparallel, line 16: Again here $\oplus$ if you care to change :)

src/algebra/README.md
```diff
@@ -0,0 +1,25 @@
...
+[Finite fields](https://en.wikipedia.org/wiki/Finite_field) are instrumental in cryptographic systems.
```
Autoparallel, line 17: Adding a touch here about how *fields* exist as special instances of rings (where there is division and no zero-divisors). 

Adding a commented TODO is also okay.

Autoparallel (approved):

src/algebra/README.md
lonerapier: I need to learn more abstract algebra :)

I think concatenation is mostly substituted as multiplication, at least in my head, so might create confusion. does using `\cdot` make more sense?

Found it used in Ben Lynn's notes as well: https://crypto.stanford.edu/pbc/notes/group/group.html

src/algebra/README.md
lonerapier: makes sense, will change it

lonerapier: done

src/algebra/README.md
lonerapier: modified it to $\cdot$, does this work?

if not can just simplify it to concatenation

lonerapier:
> One thing we could also consider is making the ring/field `+` op be a `AbelianGroup`. Food for thought.

Regarding this suggestion, I can't find a way to add this without complicating the traits and structs more

Like it wouldn't be easy for a beginner to implement 5 traits just to have a PrimeField implementation

Autoparallel:
> Like it wouldn't be easy for a beginner to implement 5 traits just to have a PrimeField implementation

That's true... it would introduce having some wrapper types and what not. The only benefit it would give is that if you did something like:
```
let element: impl FiniteField = ...;
let (element as impl AbelianGroup) + ... 
```
but that's probably not worth it.

Another option is something like
```
trait FiniteField {
    type Monoid: ...
    type AbelianGroup: ...
    ...
}
```
but then I think we are hitting abstraction for the sake of abstraction. 

You're right, let's not make the implementation needs abusive to beginners.

src/algebra/README.md
Autoparallel: $\cdot$ is fine. For a group, we just don't want to convey the operation is + as that implies commutativity in math people's heads. Concatenation is actually really general as you are thinking of group elements as strings of chars with rules in how to reduce the strings into other strings. Strings of minimal length given the reduction rules are considered canonical representatives (this, at least, is the way [group presentations](https://en.wikipedia.org/wiki/Presentation_of_a_group) define a group). All groups can be given via a presentation, albeit that they may not be finite presentations (that is really not a problem for us lol)

examples/permutation_group.rs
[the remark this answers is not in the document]
Autoparallel: Though i guess in this case D6 is S3

src/algebra/group/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+      elem = Self::op(&elem, self);
+      order += 1;
+    }
+    order
+  }
+}
+
+/// Defines a group with commutative operation: `a·b=b·a`
+pub trait AbelianGroup: Group {
+  /// Returns whether the group is an abelian group
```
Autoparallel, line 62: A little misleading. This returns true if a pair of elements commutes. 

Sadly an (almost) impossible task for Rust to handle is that there exists a proof that your impl of this trait is actually abelian (rip)

Autoparallel (approved):
Looks fantastic!

lonerapier:
> That's true... it would introduce having some wrapper types and what not.

Maybe we can implement this as an example demonstration with Ring implementation. Like showing a ring with inverse and commutativity forms a field, and then comparing it with the actual Field implementation.

src/algebra/group/mod.rs
lonerapier: Yeah, IDK why I put it this way. Have changed this to an assertion

examples/permutation_group.rs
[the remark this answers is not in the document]
lonerapier: Have changed to `symmetric_group`

PR #157, opened 2024-09-18
Add AES-GCM(+GHASH)

mrdaybird:
*fixes/closes issue #140 (Bounty: AES-GCM)*

Worklist:
- [x] Implement GHASH
- [ ] Add documentation
- [ ] Implement GCM
- [ ] Add documentation
- [ ] (optional) GHASH : faster polynomial multiplication

PR #158, opened 2024-09-23
Update README.md for modes of operation

mrdaybird:
follow up from #157 

Added more explanation for some modes, added figures for each mode, and fixed the grammar/punctuation where required.

Linked modes/README.md to root README.md

Not comprehension, but looks much better now!

@0xJepsen @brunny-eth


mrdaybird:
Unfortunately rust does not allow utf-16 in our the ascii diagrams, it looked much better with it! :cry: @0xJepsen

0xJepsen (approved):
ah thats okay this will have to do

PR #167, opened 2024-12-18
repo: create `labels.yml`

Autoparallel:
This PR adds a `labels.yml` file to our `.github` directory which will sync labels from there (though it may not delete old labels, which we may have to do manually). I got these labels from https://github.com/autoparallel/learner

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,63 @@
+- name: "bug 🐛"
+  color: "d73a4a"
+  description: "Something isn't working"
+
+- name: "documentation 📜"
+  color: "0075ca"
+  description: "Improvements or additions to documentation"
+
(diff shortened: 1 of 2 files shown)


Autoparallel:
A note for the Pluto gang -- I prefer having labels synced like this. We should consider it on all our repos and template them moving forward :)

Autoparallel:
Before merging, we should add any of the ronkathon specific labels we already have in here and want to keep.

A good example would be the "bounty" label.

0xJepsen (approved):
I like these I can't think of any on the top of my head that are missing here.

Autoparallel:
Don't merge yet -- I will merge when ready then.

Let me add some that we have that aren't currently in this yml

PR #179, opened 2024-12-26
feat: ed25519 digital signature scheme

### LINKED ISSUE #78: bounty: Edwards-curve Digital Signature Algorithm (EdDSA)
**Bounty description**

Implement the Edwards-curve Digital Signature Algorithm (EdDSA) for ronkathon. EdDSA is a modern digital signature scheme, and this implementation should focus on the Edwards25519 curve variant.

**Implementation requirements**

- [ ] Implement key generation for Ed25519 curve
    - Generate public / private key pairs
    - Ensure proper key formatting and encoding 
- [ ] Implement the EdDSA signing function, following the EdDSA specification for Ed25519
    - Implement proper message preprocessing
    - Generate valid signatures
- [ ] Implement the EdDSA sig verification function
- [ ] Implement proper error handling and input validation
- [ ] Create comprehensive unit tests, including test vectors from the EdDSA specification; 
    - Test various message lengths and edge cases
     
**Bonus features**

- [ ] Add a benchmarking suite to test performance
- [ ] Implement batch signature verification for improved efficiency

**Resources**

[RFC 8032: Edwards-Curve Digital Signature Algorithm (EdDSA)](https://datatracker.ietf.org/doc/html/rfc8032)
[Curve25519: new Diffie-Hellman speed records](https://cr.yp.to/ecdh/curve25519-20060209.pdf)
[EdDSA for more curves](https://eprint.iacr.org/2015/677.pdf)
[The Ed25519 signature scheme](https://ed25519.cr.yp.to/)

**Criteria**

Bounties will be rewarded based on the following criteria:

1. Correctness and security: A thorough review of the implementation should convince our team that they are correct and secure, with all requirements met.
2. Code clarity and quality: Succinct, easy-to-follow code with appropriate naming conventions. Utilize Rust’s type system for flexibility and security (e.g., compile-time checks where possible), and avoid external crates. Optimizations should be a lower priority than clarity, but can be included behind a feature flag as a bonus.
3. Documentation quality: Provide comprehensive README’s, Cargo docs, and inline comments where code itself is not self-explanatory. Prioritize clarity and readability.

### LINKED ISSUE #172: docs: repository MDBook?
## Idea
We have great docs located within modules, but it may be nice to also have these items brought into an MDBook so that this could be published to its own GitHub pages.

## Requirements
- [ ] Source docs from the READMEs we have so that we have these in sync with one another. 
- [ ] Set up GitHub pages deployment.

mrdaybird:
closes #78 
closes #172

- [x] Implement `keygen`, `sign`, `verify` for ed25519 scheme
- [x] add doc + better comments
- [x] add benchmark 
- [x] added sha512 
- [x] add gh-page 

Note:
1. Even though I did not want to, I have had to add a dependency([crypto-bigint](https://docs.rs/crypto-bigint/0.6.0-rc.6/crypto_bigint/index.html)) to deal with large numbers (256, 512-bit) used in the code.
2. I have also added a dev-dependency, [hex_literal](https://docs.rs/hex-literal/latest/hex_literal/), which is used for compile-time hex string to bytes conversion, which is not important but nice to have since there are a lot of such conversion throughout the repo.


mrdaybird:
@0xJepsen @Autoparallel 
I think this is ready for review!

Autoparallel:
I don't mind having either dependency -- especially dev-deps.

At some point it would be kinda cool to make our own bigint implementation. I honestly don't know the intricacies of how the popular libraries do it (i just imagine you'd do some generic ripple carry or something).

Let me review!

Autoparallel:
I think we need semver CI running on PRs into main... I can add this elsewhere and rebase here.

Cargo.toml
```diff
@@ -6,21 +6,23 @@ license    ="MIT OR Apache-2.0"
...
 
 [dependencies]
-rand     ="0.8"
-itertools="0.13"
-hex      ="0.4"
+rand          ="0.8"
+itertools     ="0.13"
+hex           ="0.4"
+crypto-bigint ="0.6.0-rc.6"
+sha2          ="0.10"
```
Autoparallel, line 16: why elevated to deps?

src/dsa/README.md
```diff
@@ -0,0 +1,50 @@
...
+**Digital Signatures** have the following properties:
+1. **Authenticity**: Just like physical signatures, digital signatures provide a way to verify the identity of a signer.
+2. **Integrity**: Digital signatures provide a mechanism to detect unauthorized modification to a message.
+3. **Non-repudiation**: Digital signatures have a nice property that once a signer signs a message, they cannot deny having done so.
+
+### How does a digital signature scheme look like?
+
+Digital signature schemes consists of three algorithms **Gen**, **Sign**, **Verify**, such that:
+
+1. The key generation algorithm, $Gen$ which takes in the security parameter $1^n$ and outputs public key, $pk$ and private key, $sk$.
```
Autoparallel, line 17: $1^n$? Or $2^n$?

Autoparallel: Or even just $n$?

src/dsa/README.md
```diff
@@ -0,0 +1,50 @@
...
+2. The signing algorithm **Sign** takes as input the keys and a message and outputs a signature.
+3. The verification algorithm **Verify**, takes as input the public key, a message, and a signature. 
+It outputs bit 1 if the signature is valid for the given message and public key, otherwise 0.
+
+### How is a digital signature scheme used?
+
+To explain how digital signature schemes are used, let's take the example of two people, Bobby and Alex.
+Bobby is the one whose signature is required, so Bobby will run the $Gen(1^n)$ algorithm to obtain, $pk, sk$.
```
Autoparallel, line 25: Nit: might want to use other tex fonts. Typically for a function like "Gen" you could do: $\textrm{Gen}$. For "pk" and "sk" you could also use something like $\texttt{pk}$ $\texttt{sk}$.

src/dsa/eddsa/curve.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Inverse of a element in the BaseField.
+/// Calculated as: x^(-1) = x^(P-2) (mod P), where P is modulus of the field.
+#[inline]
+fn inv(x: BaseField) -> BaseField { x.pow(&P_2) }
+
+/// Find the square root of an element of the `BaseField`
+/// It uses the algorithm given in Section 5.1.1 of [RFC8032] utilizing the special case of
+/// `P = 5 (mod 8)`. To read more, see: (https://en.wikipedia.org/wiki/Quadratic_residue#Prime_or_prime_power_modulus)
+pub fn sqrt(x: &BaseField) -> Option<BaseField> {
```
Autoparallel, line 96: Don't we already have a finite field `sqrt` we can use?

src/dsa/eddsa/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+use curve::{Coordinate, ScalarField, ScalarField64, GENERATOR, ORDER};
+use rand::Rng;
+use sha2::{Digest, Sha512};
+
+pub mod curve;
+
+#[cfg(test)] mod tests;
+
+/// Calculates the 64-byte SHA-512 hash.
+fn sha512_hash(bytes: &[u8]) -> [u8; 64] {
```
Autoparallel, line 15: Ah, I see where you bring this in now.

We have SHA-256. Perhaps we could just implement SHA-512 and not have this extra dep.

src/dsa/eddsa/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+    assert_eq!(public_key, expected_public_key);
+
+    let signature = Ed25519::sign(secret_key, public_key, &msg);
+    assert_eq!(signature, expected_signature);
+
+    assert!(Ed25519::verify(public_key, &msg, signature));
+  }
+}
+
+#[bench]
```
Autoparallel, line 104: Oh! Nice!

Autoparallel (commented):
This is looking great! I left comments throughout.

Mostly, I think my concern here is that this specific implementation moves further away from "first principles" than I'd like it to. Perhaps we can work this a bit to use more of ronk's native types we have (e.g., our SHA impl, PrimeField, and Curves). There would be a good amount of work to make that possible I'd imagine. I roughly see what it is, but it would be good to work towards this.

So some options are to:
- See how much we can reduce external dependency usage now with minimal pain
- Make issues for what is needed so that we move towards only using our internal types in the future.

mrdaybird:
Thanks for the review! 

> Mostly, I think my concern here is that this specific implementation moves further away from "first principles" than I'd like it to. Perhaps we can work this a bit to use more of ronk's native types we have (e.g., our SHA impl, PrimeField, and Curves). There would be a good amount of work to make that possible I'd imagine. I roughly see what it is, but it would be good to work towards this.

I agree with you completely! That was my concern as well. There are few things I would to point out.
Firstly, the issue with `PrimeField` is that it is not designed to work with large numbers, that is not just having bigint support but also having support for fast modular arithmetic using Montgomery form or Barrett reduction. (like you said, it will be cool to have our own implementation)

Secondly, the issue with ronkation's curve implementation is that Edwards curve is a different form of elliptic curve (different from the Weierstrass from that is used in ronkathon), so RFC mentions different parameters. But I think the main difference is point representation that we have used. I have had used extended homogeneous coordinates (which is the also recommended as per the RFC) because using affine coordinates was at least an order of magnitude slower. So they use different point addition formula.

Lastly, with SHA, we had SHA256 not SHA512. So I had to add a dependency. But, like you said, it will be better to implement sha512. (maybe not a lot of work, i guess)

src/dsa/eddsa/mod.rs
mrdaybird: sure, i will look into this!

src/dsa/README.md
mrdaybird: You're right, that looks better!

Cargo.toml
mrdaybird: reverted to dev-deps.

src/dsa/eddsa/mod.rs
mrdaybird: done!

src/dsa/eddsa/curve.rs
mrdaybird: If you mean the one for `PrimeField`, then I am not sure how, because this one is specialized for `BaseField` that is a type specialization of `ConstMontyForm`(from `crypto_bigint`) type

mrdaybird:
Added a sha512 implementation. I have reused most of the code from `sha256.rs`.

mrdaybird:
>So some options are to:
    - See how much we can reduce external dependency usage now with minimal pain
    - Make issues for what is needed so that we move towards only using our internal types in the future.

I think we can start with adding support for large FiniteField(and friends) which uses Montgomery reduction or Barrett reduction. I will open an issue in this regard.

src/dsa/eddsa/curve.rs
lonerapier, line 1: This is really nice, but I'd like us to add support for Edwards curve in the `curve` module. will leave the decision for adding the support for this in current PR to you (i support adding it), but we should definitely create an issue, and remove this file from `dsa` module entirely.

src/hashes/sha512.rs
lonerapier, line 1: I feel like we can combine sha256 and sha512 using const generics easily? most of the logic is duplicated

src/dsa/README.md
```diff
@@ -0,0 +1,51 @@
...
+### How is a digital signature scheme used?
+
+To explain how digital signature schemes are used, let's take the example of two people, Bobby and Alex.
+Bobby is the one whose signature is required, so Bobby will run the $\text{Gen(n)}$ algorithm to obtain, $\text{pk, sk}$. 
+Then, the public key, $\text{pk}$, is publicized as belonging to Bobby. This not only provides authentication but also ensures non-repudiation. This one of the critical parts of a secure digital signature scheme. 
+You can read more on this here: [Public key infrastructure](https://en.wikipedia.org/wiki/Public_key_infrastructure)
+
+![](./keygen.gif)
+
+Now when Alex sends a message(document, contract, etc.), $m$, for Bobby to sign, they compute the signature, $s$ as, $s\leftarrow\text{Sign(sk,m}** and sents $s$ to Alex or any other party who wants to take a look.
```
lonerapier, line 31: latex looks incorrect

lonerapier (commented):
Awesome work sir. I have few questions that I mentioned on the respective files. I think we can generalise and reduce code dedeuplication.

Adding Edwards curve support will be hugee (if not, we should create an issue)

src/hashes/sha512.rs
mrdaybird: The main issue is the different definition of sigma functions for sha512 and sha256.

src/lib.rs
```diff
@@ -21,13 +21,14 @@
 #![feature(const_option)]
 #![feature(generic_const_exprs)]
 #![feature(specialization)]
+#![feature(test)]
```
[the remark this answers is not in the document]
mrdaybird: for the benchmarks in eddsa tests.

src/dsa/README.md
mrdaybird: missed this, thanks!

src/dsa/sign_and_verify.gif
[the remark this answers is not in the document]
mrdaybird: sure!

src/hashes/sha512.rs
mrdaybird: I am a beginner in rust, so if you want you can go ahead and do that!

src/dsa/eddsa/mod.rs
```diff
@@ -0,0 +1,174 @@
+//! Edwards-curve Digital Signature Algorithm (EdDSA) based on the RFC 8032.
```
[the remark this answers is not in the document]
mrdaybird: sure!

mrdaybird:
After a bit of thought, there are two phases to going dependency free, first add bigint library and then refactor exisiting code. 
I will start by creating the initial scaffolding of bigint and a new bigint based field and curve. After that, major part will be refactoring the existing code to support bigint, for which I will create an issue.

Autoparallel:
Adding Edwards curve would maybe not be that bad? Though it would be monumental for this PR to contain it. It's a different way of writing the equation (actually a slightly different equation at that) and homogeneous coordinates are just a different coordinate system. 

It would be possible to make our curve trait generic over coordinate system and type. E.g., have something like:
```rust
EllipticCurve<Edwards, Homogeneous>
EllipticCurve<Weierstrass, Standard>
```
The main differences come down to how you implement the group operation.

---

As for the `PrimeField` concern. I'm curious if we could also put a generic in here too to handle cases for large primes. E.g., 
```rust
PrimeField<BigInt, 21888242871839275222246405745257275088548364400416034343698204186575808495617>
```
or something. In this case, we could implement the optimizations on the `BigInt` version of the `PrimeField` type. Though, I am not sure the optimizations are actually necessary as speed isn't exactly what we're going for (but we also don't want this to be incredibly slow either). I do actually think implementing optimizations is a good thing because they're a real part of most codebases. With that said, I'd be in favor of special implementations for `BigInt` type `PrimeField` elements. Specialization on structs in Rust works well at least (can't say the same for traits).

---

Also, nice work with the SHA-512. That was a good call and I appreciate you doing it. I think at this point moving forward to merge this and create issues to come back and solve these propositions would be my advice!

mrdaybird:
At some point, we have to implement our own stuff and get rid of the deps, so better now than later.
I have already started implementing `BigInt` and `BigInt` based `PrimeField` in separate modules. 
I am focusing on a "good enough" and simple implementation rather than the most optimal. This will give us a good start and later on we can improve on it. 

Another motivation for implementing it here is that I found an excellent material on it in the Handbook of Applied Cryptography (See [Chapter 14](https://cacr.uwaterloo.ca/hac/about/chap14.pdf)).

Autoparallel:
@mrdaybird I'd be very curious to review. What's your current plan for these implementations?

mrdaybird:
@Autoparallel 

At the moment, `PrimeField` struct looks something like this:
```
pub trait Modulus<T> {
    const M: T;
    fn reduce(x: Concat<T>) -> T;
}

pub struct PrimeField<T, P: Modulus<T>> {
    pub(crate) value: T,
    phantom: PhantomData<P>,
}
```
(perhaps is `PhantomData` not needed...)

Then we provide macros to create struct that implements `Modulus` trait, like:

```
create_large_modulus!(P0, BigInt<4>, /* some large hex string*/);
```

This creates a struct name `P0` that implement `Modulus<BigInt<4>>`.
We could have other macros like `create_modulus` that implements a `usize` based modulus.

lonerapier:
while i like the direction you're going, i think it will be overkill for this PR to have a BigInt implementation. We can discuss this in a separate issue.

what's your view on choosing `crypto-bigint` as the default bigint dep, and move current Field, Curve implementation to that (maybe in a separate PR)?

@Autoparallel would like your take as well.

Autoparallel:
I also think that this PR becomes unwieldy to have this much in it. 

Would also rather have it be another contribution!

I also am curious about just using `bigint` as opposed to `crypto-bigint`

src/dsa/eddsa/mod.rs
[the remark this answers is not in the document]
mrdaybird: I will leave this to you, because I am not sure how will it look like considering the ecdsa code. 
Hope that is not an issue!
Thanks!

mrdaybird:
No problem!
I am almost done with my own bigint + field implmentation. I will create a seperate PR for that.

About this PR, I have resolved/fixed all review comments as far as I understand.

Thanks!
@Autoparallel @lonerapier

mrdaybird:
> what's your view on choosing crypto-bigint as the default bigint dep, and move current Field, Curve implementation to that (maybe in a separate PR)?

If you had asked me this question before I created my own bigint + field implementation, I have preferred it. But, now I would say that it will be better to have our implementation keeping it line with ronkathon's goal of "...everything from first principles."
Also,I think this way we could have more flexibility with regard to implementation detail that is not provided by other libraries.

> I also am curious about just using bigint as opposed to crypto-bigint

After a cursory look at other bigint library([1](https://docs.rs/num-bigint/latest/num_bigint/), [2](https://docs.rs/num/latest/num/bigint/index.html)), I believe they do not provide some sort of modular reduction/multiplication support. If we want to use it, we have to write our own modular reduction/multiplication. But I am not sure if it will be better than having our own bigint impl.

@Autoparallel @lonerapier

mrdaybird:
@Autoparallel 
gentle reminder, maybe we could merge this.

Also, I have set up mdbook and added the workflow for github pages. I think it should work if you enable github pages using github actions.

Autoparallel:
@mrdaybird you might be cooking too hard!!!

Amazing you wrote a book for us to go off of too. Let me review in detail. Give me a day or so. Please be patient! I haven't forgotten you :)

book.toml
```diff
@@ -0,0 +1,18 @@
+[book]
+authors = ["Vaibhav Pathak"]
```
lonerapier, line 2: can we change the authors to ronkathon contributors or something?

lonerapier (approved):
Great work on this sir!!

book.toml
mrdaybird: Missed this! Thanks

lonerapier:
also can we create issues for adding bigint implementation, and moving all structures to bigint? maybe a bigint tracking issue

lonerapier:
can you also change eddsa in readme from in progress to done

mrdaybird:
@lonerapier 
Created an issue for bigint: #184 

Also, thanks for the review!

Cargo.toml
```diff
@@ -5,13 +5,15 @@ edition    ="2021"
 license    ="MIT OR Apache-2.0"
 name       ="ronkathon"
 repository ="https://github.com/pluto/ronkathon"
-version    = "0.1.1"
-exclude    =["CHANGELOG.md", "src/tree/ConstructMerkleTree.gif"]
+version    ="0.1.1"
+exclude    =["CHANGELOG.md", "assets/"]
```
Autoparallel, line 9: Good call

Cargo.toml
```diff
@@ -5,13 +5,15 @@ edition    ="2021"
...
 
 [dependencies]
-rand     ="0.8"
-itertools="0.14"
-hex      ="0.4"
+rand          ="0.8"
+itertools     ="0.14"
+hex           ="0.4"
+crypto-bigint ="0.6.0-rc.6"
```
Autoparallel, line 15: Saw you made issue to remove this dep 👍

README.md
```diff
@@ -34,8 +34,8 @@ to run example code.
 
 ## Primitives
 
-- [Finite Group](src/field/group.rs)
-- [Fields and Their Extensions](src/field/README.md)
+- [Finite Group](src/algebra/group/README.md)
```
Autoparallel, line 37: Should this actually read as "Finite Group" and not just "Group"?

I know you didn't make it originally, but it seems like it's generic enough. Either that or it should probably also have "Finite Fields..."

Autoparallel: This is just a nit btw

SUMMARY.md
Autoparallel, line 1: Amazing!!! It's so great with mdbook that you can just source the files from where they're already at in the repo. bullish

assets/sign_and_verify.gif
Autoparallel, line 1: Minus points for this not being Alice and Bob

(joking, but only slightly 😉)

book.toml
```diff
@@ -0,0 +1,20 @@
+[book]
+authors = ["Contributors to Ronkathon"]
+language = "en"
+multilingual = false
+src = "book"
+title = "Ronkathon Docs"
```
Autoparallel, line 6: Should we spice this up a bit? 

`title = "Ronkathon: Cryptography Educational Foundations"`

then 

`description = "Minimal, open source, education-driven cryptographic library built by community for community"`

I'm not married to this at all, but I think we can make it more exciting than just "Ronkathon Docs" 😆

Autoparallel: Especially with your manim + gifs -- this is S-tier content

src/bin/create_mdbook.rs
Autoparallel, line 1: Ahh, okay so this is how you make it pull from `src/`?

src/dsa/eddsa/curve.rs
Autoparallel: Agreed

src/dsa/eddsa/curve.rs
Autoparallel: I did mean for `PrimeField` yeah. 

Fundamentally it shouldn't matter how the field is represented though as the `sqrt` for `PrimeField` is generic

src/dsa/eddsa/mod.rs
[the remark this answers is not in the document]
Autoparallel: We should create better traits in general for the encryption module. I think we could then do a refactor there and also make a `Signature` trait.

src/dsa/eddsa/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+  // ScalarField64 calculates the modulus of 64-byte/512-bit number with `L`(the order of the
+  // subgroup)
+  let r1 = ScalarField64::new(&U512::from_le_bytes(x)).retrieve().to_le_bytes();
+  let mut r2 = [0u8; 32];
+  r2.copy_from_slice(&r1[..32]);
+  r2
+}
+
+/// Encapsulates the functions of [`Ed25519`] digital signature scheme.
+pub struct Ed25519;
```
Autoparallel, line 39: I'm in general not a fan of zero sized structs, but I know they're elsewhere in this crate too. I'd love to get rid of them. The reason why is because you could just identify the methods on `Ed25519` equivalently by the module itself so that you can just do `eddsa::keygen()` instead of `eddsa::Ed25519::keygen()`

With that said, it's as if `Ed25519` should own its private key and public key.

---

Just a nitpick here, but we have now used the terminology of "private key" and "secret key" (see the README added)

I don't mind either one, but we should pick one and stick with it. I lean slightly towards "secret key" because you can write sk/pk.

src/dsa/eddsa/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+  /// Sign the `message` using the `public_key` and `private_key`.
+  ///
+  /// It uses the algorithm given in Section 5.1.6 of [RFC8032], which is as follows:
+  ///       Notation: H <- SHA-512 hash function,
+  ///                 Enc <- Encoding function for a point on curve. See `curve.rs`
+  ///
+  /// 1. `h = H(private_key)`, 64-byte hash of private_key
+  /// 2. Split `h` into two 32-byte halves, `s` and `prefix`. `s` is used as a scalar.
+  /// 3. r = H(prefix | message), hash of prefix concatenated with the message. `r` is used as a
```
Autoparallel, line 75: I see `H` is SHA-512, but it's surprising to me that this isn't a hash to group. Is it a hash to group when combined with `Enc`?

src/hashes/sha512.rs
Autoparallel, line 1: We could probably make these generic, huh? That is something like a `SHA<256>` and `SHA<512>`

Autoparallel (approved):
Looks great @mrdaybird! I'm super excited to merge this.

I left comments, please go through and address them as you please. Resolve them along the way and let me know when you are done working on this. Then I will merge.

README.md
mrdaybird: I was wondering the same. I will improve this section a little!

assets/sign_and_verify.gif
mrdaybird: Haha! I was having a hard time drawing Alice :stuck_out_tongue_closed_eyes:! It was my first time trying on pixel art!

src/dsa/eddsa/mod.rs
mrdaybird: It is indeed hash to group, in step 1, we hash then split the hash into two halves, and the second half is used an element of the scalar field. And in step 3 the entire hash is used as an element of the scalar field. When some bytes are used as an element of the scalar field, I mention that it is `...used as a scalar`. 

I will change it to: `s is converted to an element of the scalar field`

src/hashes/sha512.rs
mrdaybird: @lonerapier also mentioned the same but the problem is that they use completely different sigma functions and K and H, plus the padding uses different constants. So I am not sure how to do this :thinking:

book.toml
mrdaybird: Definitely there is good content throughout!
Like the beautiful latex in curve and field mod. (there many such examples)
But some work needed to clean up the edges, like the symmetric encryption docs.
In the meanwhile, I will change the title and the description.

mrdaybird: I am excited to see the mdbook live!

src/bin/create_mdbook.rs
mrdaybird: Yup! I guess I will add a section on how to build/serve the mdbook locally after we go live!

src/dsa/eddsa/mod.rs
mrdaybird: Well, I agree on the fact that zero-sized structs are unnatural. 

That being said, we need it here because EdDSA have two algorithms, Ed25519 and Ed448. If someone implements Ed448, using zero-sized structs, they could very well add it inside `eddsa/mod.rs` without creating another module. 

If the user wants to use Ed25519, they could import just the `Ed25519` struct and then do `Ed25519::keygen()` etc.

-----

About the `private_key` vs `secret_key`. I will change it `secret_key`! Thanks for the suggestion!

src/hashes/sha512.rs
Autoparallel: You can have consts associated to structs based on the const generic :)

Either by a `const fn` or just a `const`

src/dsa/eddsa/mod.rs
Autoparallel: On the zero sized struct -- still could have it own the keys? Then it isn't zero sized and the methods can take in `&self` instead of taking in keys?

Autoparallel:
Yeah I'll have to enable the pages won't i...

Do you know if I should do that before or after? Or with the workflow dispatch it might not matter?

src/dsa/eddsa/mod.rs
mrdaybird: Done!

src/hashes/sha512.rs
mrdaybird: Finally got it working! It was tricky :sweat_smile: 
I have got it to work on my rust skills.

mrdaybird:
@Autoparallel 
I don't think it matters. But I tried enabling it beforehand and it worked.

src/hashes/sha512.rs
mrdaybird: I had to add `num_traits` because SHA-256 and SHA-512 use different primitive types, u32 and u64 respectively.
Otherwise, things would have gotten more complicated....

Autoparallel:
Is this good to merge on your end @mrdaybird ?

mrdaybird:
@Autoparallel I guess so...
Unless you think there is a problem.

Issue #173, opened 2024-12-19
feat: lattice-based (asymmetric) encryption scheme


Autoparallel:
## Idea
I would love to see an implementation of some NIST approved lattice based crypto scheme. The go-to example is:
- [CRYSTALS-Kyber](https://pq-crystals.org/kyber/index.shtml)

You can read more about the four approved schemes [here](https://www.nist.gov/news-events/news/2022/07/nist-announces-first-four-quantum-resistant-cryptographic-algorithms).

PR #177, opened 2024-12-23
(WIP) feat: LWE

Autoparallel:
## WIP
- [x] Basic LWE

PR #187, opened 2025-01-20
feat: Module-Lattice KEM

lonerapier:
This PR fixes/closes issue #173 

It changes the following:
- adds [ML-KEM](https://doi.org/10.6028/NIST.FIPS.203.ipd)


Autoparallel:
Should we just tag team this, add cyclotomic rings, and get this in? I can (or you can) close #177 if that's the case.

Issue #197, opened 2025-01-29
Refactoring of modules  and updating of documentation


mahmudsudo:
The present documentation does not fully cover the codebase and some modules need refactoring

Autoparallel:
@mahmudsudo want to get assigned?

If so, can you present a bit of a plan for specifics you want to accomplish? 

Agreed on this direction though! Thanks for making the issue 👍

mahmudsudo:
```
/// Core encryption trait that defines basic encryption/decryption operations
pub trait Encryption {
    /// The key type used by this encryption algorithm
    type Key;
    /// The error type returned by encryption operations
    type Error;

    /// Encrypt data using the provided key
    fn encrypt(&self, key: &Self::Key, data: &[u8]) -> Result<Vec<u8>, Self::Error>;
    
    /// Decrypt data using the provided key
    fn decrypt(&self, key: &Self::Key, data: &[u8]) -> Result<Vec<u8>, Self::Error>;
}

/// Trait for block ciphers specifically
pub trait BlockCipher: Encryption {
    /// The block size in bytes
    const BLOCK_SIZE: usize;
    
    /// Encrypt a single block
    fn encrypt_block(&self, key: &Self::Key, block: &[u8; Self::BLOCK_SIZE]) -> Result<[u8; Self::BLOCK_SIZE], Self::Error>;
    
    /// Decrypt a single block
    fn decrypt_block(&self, key: &Self::Key, block: &[u8; Self::BLOCK_SIZE]) -> Result<[u8; Self::BLOCK_SIZE], Self::Error>;
}

/// Trait for stream ciphers
pub trait StreamCipher: Encryption {
    /// The nonce type used by this cipher
    type Nonce;
    
    /// Initialize cipher state with key and nonce
    fn init(&mut self, key: &Self::Key, nonce: &Self::Nonce) -> Result<(), Self::Error>;
    
    /// Process data stream
    fn process(&mut self, data: &[u8]) -> Result<Vec<u8>, Self::Error>;
}
```


so aes and others would implement all these seperately: 
```
pub struct Aes256 {
    // Internal state
}

impl Encryption for Aes256 {
    type Key = [u8; 32];  // 256-bit key
    type Error = AesError;

    fn encrypt(&self, key: &Self::Key, data: &[u8]) -> Result<Vec<u8>, Self::Error> {
        // Implement padding and block mode handling
    }

[...]

PR #200, opened 2025-01-30
symmetric refactorings

mahmudsudo:
This PR closes issue #197 

It changes the following:
- implements the encryption trait for all symmetric ciphers
- includes the rest of the codebase in the documentation


src/encryption/mod.rs
```diff
@@ -1,3 +1,37 @@
...
+pub trait BlockOperations: Encryption {
+  /// The size of each block in bytes
+  const BLOCK_SIZE: usize;
+  /// The block type used in block-specific operations
+  type Block: AsRef<[u8]> + AsMut<[u8]> + From<Vec<u8>> + Copy + PartialEq;
+  /// Encrypt a single block of data
+  fn encrypt_block(&self, block: Self::Block) -> Result<Self::Block, Self::Error>;
+   /// Decrypt a single block of data
+  fn decrypt_block(&self, block: Self::Block) -> Result<Self::Block, Self::Error>;
+}
```
Autoparallel, line 35: On first pass I'm having trouble seeing why this is necessary, but let me get further into this review.

src/encryption/symmetric/aes/mod.rs
```diff
@@ -148,7 +84,10 @@ const ROUND_CONSTANTS: [[u8; 4]; 10] = [
 
 /// A struct containing an instance of an AES encryption/decryption.
 #[derive(Clone)]
-pub struct AES<const N: usize> {}
+pub struct AES<const N: usize>
+where [(); N / 8]: {
+  key: Key<N>,
```
Autoparallel, line 89: I'm curious if we need this `Key` wrapper struct at all anymore. Why not just have the `[u8; N / 8]` directly here? This could be more code to remove! 😄

src/encryption/symmetric/aes/mod.rs
```diff
@@ -495,21 +434,94 @@ where [(); N / 8]:
...
+  fn encrypt_block(&self, block: Block) -> Result<Block, Self::Error> {
+    self.encrypt_internal(block)
   }
 
-  fn decrypt_block(key: &Self::Key, ciphertext: &Self::Block) -> Self::Block {
-    Self::decrypt(key, ciphertext)
+  fn decrypt_block(&self, block: Block) -> Result<Block, Self::Error> {
+    self.decrypt_internal(block)
   }
 }
```
Autoparallel, line 527: Is this necessary? What does this provide?

src/encryption/symmetric/aes/mod.rs
```diff
@@ -495,21 +434,94 @@ where [(); N / 8]:
...
+  /// ```
+  fn encrypt(&self, data: &Self::Plaintext) -> Result<Self::Ciphertext, Self::Error> {
+    let num_rounds = match N {
+      128 => 10,
+      192 => 12,
+      256 => 14,
+      _ => return Err("AES only supports key sizes 128, 192 and 256 bits".to_string()),
+    };
+
+    Ok(Self::aes_encrypt(&data.0, &self.key, num_rounds))
```
Autoparallel, line 479: Here and in `decrypt` impl, I think I'd rather just have the implementation be here rather than indirectly call into `Self::aes_encrypt`. This makes reading the code easier imo. What do you think?

src/encryption/symmetric/des/mod.rs
```diff
@@ -21,7 +21,10 @@ pub type SubKeys = [Subkey; 16];
 
 /// DES encryption
 #[derive(Debug)]
-pub struct DES {}
+pub struct DES {
+  key:     Block,
+  subkeys: SubKeys,
```
Autoparallel, line 26: Do you prefer the type aliases that we've been using here? I personally don't find them helpful, but maybe others do. If you don't think they add anything, we could delete them and do it more directly here?

My opinion on type aliases is that they should be for complicated types and for complicated types that you reuse throughout the codebase. These are not complicated (imo) and are only used here.

src/encryption/symmetric/des/mod.rs
```diff
@@ -223,44 +226,46 @@ impl DES {
...
   ///
-  /// let subkeys = DES::setup(secret_key);
-  ///
   /// let message = rng.gen();
   /// let encrypted = DES::encrypt(&subkeys, &message);
   /// ```
-  fn encrypt(key: &Self::Key, plaintext: &Self::Block) -> Self::Block {
-    Self::des(plaintext, key.iter())
+  fn encrypt(&self, data: &Self::Plaintext) -> Result<Self::Ciphertext, Self::Error> {
+    Ok(Self::des(data, &mut self.subkeys.iter()))
```
Autoparallel, line 252: The indirection here again is not my favorite. It's odd to call into a function also just called `des`. Perhaps just move the implementation into here directly? I think this is the case for all our encryption implementations.

src/encryption/symmetric/modes/gcm.rs
```diff
@@ -45,12 +45,12 @@ use crate::{
 ///
 /// # Generics
 /// - `C`: A block cipher that implements the `BlockCipher` trait.
-pub struct GCM<C: BlockCipher> {
+pub struct GCM<C: BlockOperations + Encryption> {
   ghash: GHASH,
```
Autoparallel, line 49: Random, but I wonder why `GHASH` is all caps? (Not your problem)

Autoparallel (commented):
Nice work so far! This is coming along really well.

I left a good amount of comments and I left some decisions to be made at your discretion.

A challenge to you: how much code and indirection can we remove? I'd love to see this PR achieve a diff that cuts down LOC. I think it's possible. The abstractions are getting better. 

Let me know if you want any collaboration here!

src/encryption/mod.rs
mahmudsudo: to provide access for encrypt_block functions as it is not used by all the ciphers 

this is relied on in the modes crate and use of generics , we could add it as part of the encryption crate but that will implementing it for ciphers that dont need it

src/encryption/symmetric/aes/mod.rs
mahmudsudo: actually it should reduce loc if it is removed , as the Key is just an abstraction , i will refactor and revert

src/encryption/symmetric/aes/mod.rs
mahmudsudo: THe circular dependency of the BlockOperations on (Encryption)itself was the reason for the use of the helper functions ,another way would be to re implement encrypt in the encrypt_block

src/encryption/symmetric/aes/mod.rs
mahmudsudo: Will refactor and resolve

src/encryption/symmetric/des/mod.rs
mahmudsudo: i didnt want to make large changes to the codebase initially 
i will refactor and resolve

src/encryption/symmetric/des/mod.rs
mahmudsudo: seems good , move the implementation  into the encrypt and reuse encrypt in decrypt

src/encryption/symmetric/aes/mod.rs
mahmudsudo: on second thought, the struct key type needs some helper functions that might introduce more loc if refactored

src/encryption/symmetric/aes/mod.rs
Autoparallel: I guess what I'm asking is if we need the `BlockOperations` trait at all. Does it not suffice to just have this implemented inside of the `encrypt`/`decrypt` methods?

Autoparallel (approved):
Looks good to me! 

@lonerapier do you want to give a quick review as well?

mahmudsudo:
> Looks good to me!
> 
> @lonerapier do you want to give a quick review as well?

the blockOperations is needed in modes

src/encryption/symmetric/aes/mod.rs
mahmudsudo: it is needed in the modes generic usage  and it isnt included in Encryption because other ciphers dont need it

examples/aes_chained_cbc.rs
```diff
@@ -64,7 +64,7 @@ fn main() {
 
   // select a uniform bit b, and chose message m_b for encryption
   let bit = rng.gen_range(0..=1);
-  let encrypted = cbc.encrypt(&key, messages[bit]);
+  let encrypted = cbc.encrypt(key, messages[bit]);
```
lonerapier, line 67: don't understand the rationale behind not using key as reference and doing a copy for each encrypt/decrypt call. Is it faster? do we have any benches to compare against?

src/compiler/README.md
```diff
@@ -112,21 +114,21 @@ pub struct CommonPreprocessedInput {
   /// multiplicative group order
   pub group_order: usize,
   /// Q_L(X): left wire selector polynomial
-  pub ql:          Poly,
+  pub ql:          usize,
```
lonerapier, line 117: I think this is a polynomial only. Can be changed to `Polynomial<Lagrange, PlutoScalarField, GROUP_ORDER>`.

Refer:  https://github.com/pluto/ronkathon/blob/main/src/compiler/program.rs#L78-L97

lonerapier (approved):
awesome cleanup. just couple of questions :)

Let's merge when they're resolved!!

examples/aes_chained_cbc.rs
mahmudsudo: just a design choice , borrows are though theoretically less efficient memory wise

lonerapier:
@mahmudsudo doc test for recent change fails

also can you write a quick benchmark to justify the new reference change?

mahmudsudo:
> @mahmudsudo doc test for recent change fails
> 
> also can you write a quick benchmark to justify the new reference change?



> @mahmudsudo doc test for recent change fails
> 
> also can you write a quick benchmark to justify the new reference change?

The reference also came up as an issue due to using generics that arent specified as copyable or clonable .
hence the changes 

the docutests have been fixed

Autoparallel:
> The reference also came up as an issue due to using generics that arent specified as copyable or clonable . hence the changes

Can you explain what you mean and point to a specific bit of code?

src/encryption/symmetric/modes/cbc.rs
```diff
@@ -59,10 +59,14 @@ impl<C: BlockCipher> CBC<C> {
       let length = C::BLOCK_SIZE - (plaintext.len() % C::BLOCK_SIZE);
       plaintext.extend(std::iter::repeat(length as u8).take(length));
     }
-
+    let value = C::new(key);
```
mahmudsudo, line 62: the new function here takes ownership of the key type and also its signature takes in the actual type not the reference

mahmudsudo: the existence of new functions brought the removal of the borrow

mahmudsudo:
it is important to note that , the whole usage of the encrypt and decrypt functions have changed from associative functions to methods hence the only times the keys are used as params is in the modes and the new function

src/encryption/symmetric/modes/cbc.rs
mahmudsudo: C::Key type isnt known until runtime as the type is from the trait ,hence the borrow , but i can still revert to the borrow if that is preferred

lonerapier:
looks good. thanks for these changes, merging it!!

PR #209, opened 2025-02-04
fix: fix incorrect elliptic curve point initialization

famouswizard:
A mistake in the way the elliptic curve point `G2` is initialized.
In SageMath, points should be defined as `E(x, y)`, not `E([x, y])`.  

**Before:**  
```python
G2 = E2([36, 31 * X])
```  
**After:**  
```python
G2 = E2(36, 31 * X)
```

### DIFF
--- a/math/curve.sage
+++ b/math/curve.sage
@@ -46,7 +46,7 @@ print(G1)
 N = 17
 
 # G2 is the generator for E2
-G2 = E2([36, 31 *X])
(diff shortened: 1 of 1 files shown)


Autoparallel:
@famouswizard I believe this is equivalent in Sage, no? Please give a working example if not

Autoparallel:
> [Autoparallel](https://github.com/Autoparallel), its Saga.

What?

PR #214, opened 2025-02-05
fix: `index.html` renaming of `README.md`

Autoparallel:
This makes links work across book and repo./

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -16,9 +16,11 @@ use-default-preprocessors=false
 [preprocessor.katex]
 after=["links"]
 
-[preprocessor.linkcheck]
+[output.linkcheck]
 follow-web-links           =true
 traverse-parent-directories=true
+warning-policy             ="ignore"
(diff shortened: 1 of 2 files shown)


mrdaybird:
`mdbook-linkcheck` does not seem to be working.

https://github.com/pluto/ronkathon/blob/38f2b1b495adb43cef80b6c95ed0f740f901d680/book.toml#L19-L21

`[preprocessor.linkcheck]` should be `[output.linkcheck]`

After this change, `mdbook-linkcheck` gives a couple of errors and a bunch of warning...

Autoparallel:
Good catch, @mrdaybird -- as always

Autoparallel:
@0xJepsen @lonerapier can we get a review and merge? I don't care that the arkworks link is broken in the test. That is just dumb.

0xJepsen (approved):
looks good to me, i whitelisted the arkworks link, thats weird it was causing a failure

Issue #82, opened 2024-05-25
feat: Shamir's secret sharing


lonerapier:
Resources:

- https://web.mit.edu/6.857/OldStuff/Fall03/ref/Shamir-HowToShareASecret.pdf
- https://decentralizedthoughts.github.io/2020-07-17-polynomial-secret-sharing-and-the-lagrange-basis/
- https://crypto.stanford.edu/cs355/19sp/lec6.pdf
- [verifiable secret sharing](https://en.wikipedia.org/wiki/Verifiable_secret_sharing)?

eightfilms:
i can work on this unless you're picking this up! @lonerapier

lonerapier:
It's all yours sir. I'd ask @Autoparallel, he was interested in tackling this

Autoparallel:
I haven't assigned this, i think this is a great problem for another contributor actually :)

lonerapier:
I think we should start assigning good first issue labels, so that it's easier for any contributor

Autoparallel:
@eightfilms it's been a while, are you interested? I can assign!

eightfilms:
hey @Autoparallel - would be more interested in tackling #116 instead. Will leave a comment there!

Autoparallel:
Noted! Enjoy!

mahmudsudo:
@Autoparallel any other review on the pr ?

mahmudsudo:
this can also be closed

PR #215, opened 2025-02-05
shamir secret sharing scheme

mahmudsudo:
This PR closes issue #82

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,7 @@ pub mod multi_var_poly;
 pub mod polynomial;
 pub mod sumcheck;
 pub mod tree;
+pub mod shamir;
 
 use core::{
   fmt::{self, Display, Formatter},


Autoparallel:
Leave some for the rest of us!!! 

Just kidding. Let me review :)

Autoparallel (changes requested):
A few comments:

This does 2/2 secret sharing. Do you want to implement k/n secret sharing? I don't think it's fundamentally that much more work to make it generic.

Can you please write up a nice README here that explains this topic deeply? We aim to have something like that for every module/submodule. This particular tool has a lot of application usecases. Perhaps you can talk about those and how this is used?

mahmudsudo:
it is presently generic , just not enforced that n = 2k-1 or floor(n/2) = k-1

src/shamir/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// # Panics
+///
+/// This function panics if `threshold` is 0, if `share_count` is less than `threshold`, or if the
+/// secret is not less than PRIME.
+///
+/// # Returns
+///
+/// A vector of shares, where each share is a tuple (x, y).
+pub fn split_secret(secret: u128, threshold: usize, share_count: usize) -> Vec<Share> {
```
mahmudsudo, line 35: this takes in a share_count , n and threshold k

src/shamir/README.md
```diff
@@ -0,0 +1,100 @@
...
+let recovered = combine_shares(required_shares);
+assert_eq!(secret, recovered);
+```
+
+
+## Conclusion
+
+Shamir's Secret Sharing Scheme is a powerful and versatile technique for enhancing security through distributed trust. Whether it is used for secure key backup, multisignature wallets, controlled access environments, or threshold cryptographic protocols, this module offers a robust, generic implementation that can be tailored to many different requirements.
+
+This README provides an in-depth overview of both the theory and practice behind the implementation, ensuring that users and developers can make full use of the module's capabilities.
```
lonerapier, line 100: this seems like LLM generated, can we rephrase it a little better. adding more introductor references, maybe removing this conclusion entirely

src/shamir/README.md
```diff
@@ -0,0 +1,100 @@
...
+  This generic approach makes the scheme highly flexible and suitable for a wide range of applications.
+
+- **Core Functions:**
+  - `split_secret(secret: u128, threshold: usize, share_count: usize) -> Vec<Share>`  
+    Splits the given secret into the specified number of shares.
+    
+  - `combine_shares(shares: &[Share]) -> u128`  
+    Reconstructs the secret from a collection of shares using Lagrange interpolation.
+
+## Applications and Use Cases
```
lonerapier, line 48: i don't think this is needed, adding references should work

src/shamir/README.md
mahmudsudo: ok on it

Autoparallel:
Weird... we have the CBC mode failing on encryption right now....

mahmudsudo:
> Weird... we have the CBC mode failing on encryption right now....

Claims index out of bounds.

Autoparallel:
@mahmudsudo this was passing with the refactor in the encryption module. It's odd to me that it failed here. 

I'm going to rerun the workflows to see if it is probabilistic.

mahmudsudo:
I think it had something to do with the rng

src/shamir/mod.rs
Autoparallel: Ah that's my bad. Thanks for clarifying

src/shamir/README.md
```diff
@@ -0,0 +1,57 @@
+# Shamir's Secret Sharing Scheme
+
+Shamir's Secret Sharing (SSS) is a powerful cryptographic technique that divides a sensitive secret into multiple parts (shares), so that the secret can only be reconstructed when a minimum number of shares (the threshold) are combined. Fewer than the required number of shares reveal no useful information about the secret.
+
+## Background and Theory
+
+At its core, Shamir's scheme relies on the mathematical principle that a polynomial of degree (k - 1) is uniquely determined by k distinct points. The idea is to embed the secret as the constant term in a random polynomial:
+
+  P(x) = a₀ + a₁·x + a₂·x² + ... + aₖ₋₁·x^(k-1)
```
Autoparallel, line 9: Use LaTeX here?

src/shamir/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here, a₀ represents the secret, and the subsequent coefficients (a₁, a₂, …, aₖ₋₁) are randomly chosen from a finite field. Each share is generated by evaluating the polynomial at a unique nonzero value of x (for example, x = 1, 2, …, n).  
+Reconstruction is achieved using Lagrange interpolation, which recovers the polynomial—and thus the secret—from k shares.
+
+## Finite Field Considerations
+
+This implementation performs arithmetic in the finite field GF(PRIME), where:
+
+  PRIME = 2^127 - 1  
+   = 170141183460469231731687303715884105727
```
Autoparallel, line 19: similarly latex

Autoparallel (commented):
My only request now is just use LaTeX for formatting in the README -- follows the rest of them.

src/shamir/README.md
mahmudsudo: On it

Autoparallel (commented):
I know it's a nit, but there are still spots where LaTeX is not being used. E.g., 

(a₁, a₂, …, aₖ₋₁)

For inline TeX, use `$ ... $` and for centered newline/displaystyle TeX use `$$...$$`

src/shamir/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+//! of shares such that any subset of shares of size at least `threshold` (k,n) where n = 2k-1
+//! can reconstruct the secret using Lagrange interpolation over a finite field.
+//!
+//! The chosen finite field is GF(PRIME) with PRIME = 2^127 - 1. Be sure that the provided
+//! secret is less than PRIME.
+
+use rand::Rng;
+
+/// Prime modulus for finite field arithmetic: 2^127 - 1.
+const PRIME: u128 = 170141183460469231731687303715884105727;
```
lonerapier, line 13: why have we chosen a specific prime, and not made it using ronkathon field implementation?

src/shamir/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// # Arguments
+///
+/// * `coeffs` - Slice of polynomial coefficients.
+/// * `x` - The point at which to evaluate the polynomial.
+///
+/// # Returns
+///
+/// The evaluated value modulo PRIME.
+fn eval_poly(coeffs: &[u128], x: u128) -> u128 {
```
lonerapier, line 106: i think these could be avoided if we used ronkathon poly?

Autoparallel:
@mahmudsudo sorry for the delay on my side -- I think @lonerapier 's feedback is good. Let's take that into account and get this merged in :)

mahmudsudo:
@lonerapier @Autoparallel

Autoparallel:
On it

Autoparallel:
@mahmudsudo this did not resolve @lonerapier's suggestion to make this generic over our `Field` implementation. Rather you used the `PLUTO_PRIME`. Please use the other modules we have available to work on this.

mahmudsudo:
@Autoparallel Field type is nonexistent , only Field trait exists which is hard to make generic due to some specific functions not in the trait but in the actual impl block , @lonerapier review only asked to use the ronkathon field calculations as opposed to implementing with an arbitrary prime number and doing all calculations by hand

mahmudsudo:
with the update i have converted every calculation and usage to be based on the field calculation , field here isnt a type but a generic word about the PrimeField

Autoparallel:
By field type I mean field trait. This should be able to be generic over the field.

mahmudsudo:
There are some specific functions that arent provided by traits that are needed , a clear example is in the evaluation of the polynomial , making it generic over the field trait is almost impossible because of the absence of specific functions which are present in the impl block of the PlutoBaseField

lonerapier:
> There are some specific functions that arent provided by traits that are needed

then let's add those functions to the trait, and make an impl available in the types. 

Utility of this module is negligible if it's instantiated on a particular type. Ronkathon is a library, and it's written in a way that's generic over the types and can be used in any context. Maybe you can create a new trait with the functions required to make secret-sharing generic.

we should also think about how to make this easily extendable to other [secret sharing schemes](https://www.esat.kuleuven.be/cosic/blog/lsss/) like Linear SS, or Additive SS, so that it's not just limited to shamir secret sharing.

There's no push for you to make these changes. if you think this is enough, i'm happy to merge current version as an [example](https://github.com/pluto/ronkathon/tree/main/examples) in the lib.

mahmudsudo:
with your explanation i will love to make the corrections .

mahmudsudo:
@Autoparallel @lonerapier

Autoparallel:
On it

src/shamir/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+To generate shares from a secret, you specify the secret (a u128 number), the threshold (k), and the total number of desired shares (n):
+
+```rust
+use ronkathon::shamir::{split_secret, combine_shares};
+
+let secret: u128 = 123456789;
+let threshold = 3;
+let share_count = 5;
+let shares = split_secret<threshold>(secret, share_count);
```
Autoparallel, line 43: nit: this wouldn't compile
```suggestion
const THRESHOLD: PrimeField = PrimeField::from(3);
let shares = split_secret::<THRESHOLD>(secret, share_count);
```

Autoparallel (changes requested):
Left one small nit, but doc example wouldn't work

src/shamir/README.md
mahmudsudo: this seems wrong , the THRESHOLD TYPE is a usize not a PRIMEFIELD


``
pub fn split_secret<F: FiniteField, const THRESHOLD: usize>(
  secret: F,
  share_count: usize,
) -> Vec<Share<F>>
``

Autoparallel: ah oops, that's my bad. Either way, you can only pass a `const` to a function like that and the `let` binding is not `const`

Autoparallel (approved):
Thank you!

It would be great to run these snippets with `mdbook-keeper` at some point.

Autoparallel:
@mahmudsudo rebasing this to fix a small diff and then will merge.

Again, good work!

Issue #195, opened 2025-01-28
feat: BLS signatures


Autoparallel:
## Bounty Description 
Implement Boneh-Lynn-Shacham (BLS) signatures for ronkathon. BLS signatures are a cryptographic signature scheme that enables signature aggregation and operates on pairing-friendly curves. This implementation will build upon ronkathon’s existing curve and pairing primitives to demonstrate signature generation and verification from first principles.

### Implementation Requirements
- Implement core BLS signature components in the `signatures/bls.rs` module:
    - Key generation using ronkathon’s existing group operations on pairing-friendly curves
    - Signing function that maps messages to curve points using a hash-to-curve function
    - Verification function using the existing pairing implementation
    - Basic signature aggregation functionality for multiple signatures
- Utilize existing ronkathon primitives:
    - Build upon the `curve/` module for elliptic curve operations
    - Use the existing pairing implementation from the `curve/pairings.rs` module
    - Leverage the hash functions from the `hashes/` module for message preprocessing
- Implement proper error handling and input validation:
    - Verify curve points are in the correct subgroup
    - Handle invalid inputs and edge cases
    - Provide clear error messages
- Documentation and tests:
    - Create comprehensive test vectors including known test cases
    - Test signature aggregation with multiple signers
    - Test edge cases like empty messages and invalid points
    - Document security assumptions and usage guidelines
    - Include examples demonstrating proper usage

### Bonus Features
- Implement batch verification for improved efficiency
- Add support for proof-of-possession to prevent rogue key attacks
- Create benchmarking suite comparing single vs aggregated verification
- Implement optimized hash-to-curve algorithms

### Resources
- [BLS Signatures Specification](https://datatracker.ietf.org/doc/draft-irtf-cfrg-bls-signature/)
- [Boneh-Lynn-Shacham Short Signatures (Original Paper)](https://iacr.org/archive/asiacrypt2001/22480516.pdf)
- [Hash to Curve Specification (RFC 9380)](https://datatracker.ietf.org/doc/rfc9380/)
- [BLS12-381 for the rest of us](https://hackmd.io/@benjaminion/bls12-381#BLS-digital-signatures)

### Integration Guidelines
The implementation should:
- Follow ronkathon’s modular design pattern
- Use traits to define signature scheme interfaces
- Build upon existing field and curve arithmetic
- Minimize external dependencies
- Prioritize clarity and educational value over optimization

### Criteria
Bounties will be rewarded based on the following criteria:
[...]

PR #216, opened 2025-02-06
 bls signature

mahmudsudo:
This PR closes issue #195


Autoparallel:
Nice work so far. Ping me or @lonerapier when a review is needed.

mahmudsudo:
@Autoparallel i think it is ready for review

Autoparallel:
Alright let me check this out

src/signatures/bls.rs
```diff
@@ -0,0 +1,646 @@
...
+  VerificationFailed,
+  /// Other error with a descriptive message.
+  Other(String),
+  /// Invalid point encountered.
+  InvalidPoint,
+}
+
+/// BLS private key.
+pub struct BlsPrivateKey {
+  sk: PlutoScalarField,
```
Autoparallel, line 49: Can this use the `Field` type we already have instead of picking a specific prime field?

src/signatures/bls.rs
```diff
@@ -0,0 +1,646 @@
...
+}
+
+/// BLS public key.
+pub struct BlsPublicKey {
+  pk: AffinePoint<PlutoBaseCurve>,
```
Autoparallel, line 54: Can this be generic over curves?

src/signatures/bls.rs
```diff
@@ -0,0 +1,646 @@
...
+}
+
+/// BLS signature.
+pub struct BlsSignature {
+  sig: AffinePoint<PlutoExtendedCurve>,
```
Autoparallel, line 59: Likewise, can this be generic over a curve over an extension?

src/signatures/bls.rs
```diff
@@ -0,0 +1,646 @@
...
+
+/// Converts a nonnegative integer to an octet string of a specified length using crypto-bigint.
+///
+/// I2OSP (Integer-to-Octet-String Primitive) converts a nonnegative integer `x` (represented as a
+/// 256-bit integer) into its big-endian representation, trimmed of any excess leading zeroes, and
+/// left-padded with zeroes so that the result has exactly `length` bytes.
+///
+/// # Arguments
+///
+/// * `x` - A reference to a `U256` representing the nonnegative integer.
```
Autoparallel, line 76: `x` is U256 based on this comment, but is a `usize` in the function signature

src/signatures/bls.rs
```diff
@@ -0,0 +1,646 @@
...
+      for byte in tv {
+        val = (val * 256 + *byte as usize) % p;
+      }
+      e_vals[j] = PrimeField::new(val);
+    }
+    result.push(PlutoBaseFieldExtension::new(e_vals));
+  }
+
+  result
+}
```
Autoparallel, line 234: This really should be using `GaloisField` and its associated const generics. 

Also, what python implementation is this referencing?

Autoparallel (changes requested):
Left some comments. Over all looking good. Can you add more documentation to the modules you created?

Also, can you add a `README.md` that explains this in detail?

src/signatures/bls.rs
mahmudsudo: PlutoBaseFieldExtension is a specific  instance of the galois field  and its the extension field of choice in the pluto ronkathon system 

`/// The [`PlutoBaseFieldExtension`] is a specific instance of the [`GaloisField`] struct with the
/// order set to the prime number `101^2`. This is the quadratic extension field over the
/// [`PlutoBaseField`] used in the Pluto `ronkathon` system.
pub type PlutoBaseFieldExtension = GaloisField<2, 101>;
`

src/signatures/bls.rs
mahmudsudo: Field type ?? , there is only a field trait not type 
`
/// A field is a set of elements on which addition, subtraction, multiplication, and division are
/// defined.

#[const_trait]
pub trait Field:
}`
and the privateKey is just a number hence why it was easier as a scalar field, but i can make it generic  on the trait Field

src/signatures/bls.rs
mahmudsudo: Generic over the Elliptic curve trait 

that would be an hurdle due to the need to convert from one curve to the other and this conversion needs access to the basefield

src/signatures/bls.rs
mahmudsudo: in the previous implementation i used the u256 type from crypto_biguint hence why that was in the comment

src/signatures/bls.rs
mahmudsudo: same issue as the public key

src/signatures/bls.rs
Autoparallel: Yes I get that, but it can be done generically, no? So you can use `GaloisField<2, P>` for some prime `P`

mahmudsudo: making it generic over `GaloisField<2, P>` has a little bottleneck , The ExtensionField trait which is responsible for most operation between same galoisfield is only implemented for the PlutoBaseFieldExtension or `GaloisField<2, 101>` whic means the genericness can only occur in the hash_to_field function as the hash_to_curve function uses specific operations that are only implemented for the PlutoBaseFieldExtension

Autoparallel: Seems like we should implement all the necessary features on `GaloisField<2,P>` if we can then. That would be a great help!

src/signatures/bls.rs
Autoparallel: Well, it would be generic over a curve, then that curve over an extension field. So not entirely generic.

src/signatures/bls.rs
Autoparallel: Type and trait are fairly interchangeable in Rust. Yes, I mean trait not type alias

mahmudsudo:
@Autoparallel

Autoparallel:
@mahmudsudo yes, sorry. I was out. Let me also ping my teammates.

src/algebra/field/extension/arithmetic.rs
```diff
@@ -155,4 +155,60 @@ impl<const N: usize, const P: usize> Mul<GaloisField<N, P>> for PrimeField<P> {
     res
   }
 }
+
+impl FieldExt for GaloisField<2, 101> {
```
Autoparallel, line 159: Stepping through this in order, but why are we only implementing on `GaloisField<2,101>` and not on `GoloisField<Q,P>`?

We should be working over generics, not implementing for specific fields!

Autoparallel: Just to be clear, I'm aware that not every possible field (nor its extension) may satisfy what we need to do BLS. But we should implement BLS for the widest class of possibilities which means constraining those by use of the type system and compile time (const) checks.

src/algebra/field/mod.rs
```diff
@@ -74,3 +74,11 @@ pub trait FiniteField: Finite + Field {
...
   }
 }
+
+/// traits that define some functions usable in signatures
+pub trait FieldExt: Sized {
+  /// Returns the square root of the field element.
+  fn sqrt(&self) -> Option<(Self, Self)>;
+
+  fn euler_criterion(&self) -> bool;
+}
```
Autoparallel, line 84: We should just place these on the `Field` trait itself, or implement them on the extension field struct (as we did with the `PrimeField` struct)

mahmudsudo: This was made a new trait because : 
1)  Field trait has some implementation that dont use this functions or doesnt provide functions to give the functionalities so also the extension trait.
2) this reduces any redundant code and tries to minimize unused implementation .

mahmudsudo:
@Autoparallel

src/algebra/field/extension/arithmetic.rs
Autoparallel: But why are we constraining to implement only on `PlutoBaseField` as well as its degree-2 extension `GaloisField<2,101>`. We should have a generic implementation for any fields/curves that fit the necessary criteria. What I'm suggesting is to have that criteria checked within the type system.

Happy to try to help there. Can you show me what is failing if you try to handled this with generic curves/fields?

mahmudsudo: The exact issue is during the pairing , there is need to convert between fields ,using generic fields here , is impractical as different fields  have different conversion hence why the present bls makes specific concession towards the two known fields codebase i.e PlutoBase and the extension , the conversion between two fields that are generic is quite an hurdle .

mahmudsudo: that implementation was implemented for plutobase and its extended field because they are the fields used for testing , the actual bls uses generics , you could check the implementation again , the only part still specific is the parts partaining to converting between fields i.e pairing and verifying the signature.
  
  an additional check on the fields can also be added

mahmudsudo: Additionally 👍 : 

`
impl Field for BinaryField {
  const ONE: Self = BinaryField::One;
  const ZERO: Self = BinaryField::Zero;

  fn inverse(&self) -> Option<Self> {
    match *self {
      Self::Zero => None,
      Self::One => Some(Self::One),
    }
  }

  fn pow(self, _: usize) -> Self { self }
}`


as can be seen from the snippet , binaryField implements the field trait so also is other fields that neither uses norms or sqrt defined in the FieldExt trait hence the FieldExt trait cant be added to the Field trait as this would break many implementations and cause unused code in other parts , additional checks for the FieldExt will be pushed

mahmudsudo:
> @mahmudsudo let me go deeper here and do a better review.
> 
> Can you resolve these current conflicts the branch has against main?

done

Autoparallel:
@mahmudsudo I did a bit of a refactor in the signatures module.

Let's do a few things:
- Can you write up a `README.md` for the `signatures` module that explains digital signatures more broadly?
- Can you add to the `bls/README.md` and expand on how signature aggregation works? This is the key component of BLS after all!

mahmudsudo:
@mahmudsudo

src/signatures/README.md
Autoparallel, line 1: This is looking good too!

src/algebra/field/mod.rs
```diff
@@ -74,3 +74,11 @@ pub trait FiniteField: Finite + Field {
     Self::PRIMITIVE_ELEMENT.pow(pow)
   }
 }
+
+/// traits that define some functions usable in signatures
+pub trait FieldExt: Sized {
```
Autoparallel, line 79: Still not a fan of adding an additional trait. But we can always refactor later.

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+///
+/// * `Ok(Vec<u8>)` containing the octet string if the integer can be represented in the specified
+///   length.
+/// * `Err(String)` if the integer is too large to be encoded in the given number of octets.
+///
+/// # Example
+///
+/// ```
+
+/// ```
```
Autoparallel, line 84: nit: there's some weird formatting like this here. Do you want to add in examples or clear this away? (Please fix in all spots)

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    // For SHA-256, the hash length is 32 bytes.
+    vec![0u8; 32]
+  } else {
+    salt.to_vec()
+  };
+  hmac_sha256(&salt, ikm).to_vec()
+}
+
+/// Implements expand_message_xmd as specified in the standard
+fn expand_message_xmd(msg: &[u8], dst: &[u8], len_in_bytes: usize) -> Vec<u8> {
```
Autoparallel, line 147: Though not `pub`, can we add more documentation here? This function is sufficiently complex to warrant some more in-depth docs.

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    hasher.update(&dst_prime);
+    uniform_bytes.extend_from_slice(&hasher.finalize());
+  }
+
+  uniform_bytes.truncate(len_in_bytes);
+  uniform_bytes
+}
+
+/// Implements hash_to_field as specified in the standard
+fn hash_to_field<C: EllipticCurve, D: EllipticCurve>(
```
Autoparallel, line 202: This also could be a nice `pub` function that could be used elsewhere. Either way, can you also add more docs here?

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        * <D as EllipticCurve>::ScalarField::from(self.sk.into()),
+    };
+    Ok(ProofOfPossession { pop })
+  }
+}
+impl<C: EllipticCurve> BlsPublicKey<C> {
+  /// Verifies a BLS signature against the given message.
+  ///
+  /// The verification check uses the bilinear pairing:
+  ///   e(signature, G) == e(H(message), public_key)
```
Autoparallel, line 306: man, pairings are so helpful!

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+    let hash_point = hash_to_curve::<C, D>(msg)?;
+
+    // Build the properly twisted generator G from the base-curve generator.
+    let g = convert_to_extended::<C, D>(AffinePoint::<C>::GENERATOR);
+
+    // Convert the public key into the extended group and canonicalize.
+    let pk = convert_to_extended::<C, D>(self.pk);
+
+    // Compute the pairing outputs.
+    let left = pairing::<D, 17>(signature.sig, g);
```
Autoparallel, line 326: 17? Should this be some associated constant to the `EllipticCurve`? What happens if we use a different curve/field?

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+{
+  if pks.is_empty() || messages.is_empty() || pks.len() != messages.len() {
+    return Err(BlsError::Other("Invalid input lengths".to_string()));
+  }
+
+  // Build the same properly twisted generator G.
+  let g = convert_to_extended::<C, D>(AffinePoint::<C>::GENERATOR);
+
+  // Verification: e(aggregated_sig, G) must equal the product over all pairings.
+  let left = pairing::<D, 17>(aggregated_sig.sig, g);
```
Autoparallel, line 395: 17

src/signatures/bls/tests.rs
Autoparallel, line 1: Can you add a few tests here that show how this can lead to efficient signature aggregation? It would also be worth comparing this to verifying multiple signatures in the docs too so that the advantage is clear.

src/signatures/bls/mod.rs
mahmudsudo: fixed

src/signatures/bls/tests.rs
mahmudsudo: there is a test for test aggregation already

mahmudsudo:
@Autoparallel

src/signatures/bls/README.md
```diff
@@ -0,0 +1,65 @@
...
+1. **Size Efficiency**: The aggregate signature has the same size as a single signature
+2. **Verification Efficiency**: Verifying the aggregate signature is faster than verifying n signatures individually
+3. **Non-interactivity**: Signers don't need to coordinate or communicate
+4. **Public Key Aggregation**: Public keys can also be aggregated
+
+### Mathematical Background
+
+BLS signatures work on special elliptic curves with bilinear pairings. The key components are:
+
+1. **Groups**: Uses two groups G₁ and G₂ with a bilinear mapping $e: G_1 × G_2 → G_t$
```
Autoparallel, line 57:
```suggestion
1. **Groups**: Uses two groups $G_1$ and $G_2$ with a bilinear mapping $e: G_1 \times G_2 \to G_t$
```

Learning tex is always a good thing

src/signatures/bls/README.md
```diff
@@ -0,0 +1,65 @@
...
+## Signature Aggregation
+
+Multiple signatures can be aggregated into a single signature, which can then be verified against the combined public keys. This is particularly useful in scenarios where multiple parties sign the same message.
+
+### Basic Principle
+Given multiple (message, signature) pairs from different signers:
+- (m₁, σ₁) from signer 1
+- (m₂, σ₂) from signer 2
+- ...
+- (mₙ, σₙ) from signer n
```
Autoparallel, line 43:
```suggestion
- ($m_1$, $\sigma_1$) from signer 1
- ($m_2$, $\sigma_2$) from signer 2
- ...
- ($m_n$, $\sigma_n$) from signer $n$
```

src/signatures/bls/README.md
Autoparallel, line 1: Can you add in here how the aggregation is actually faster to verify than verifying many signatures individually? Furthermore, it would be nice to have a test for it too.

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+      e_vals[j] = <C as EllipticCurve>::BaseField::from(val);
+    }
+    result.push(<D as EllipticCurve>::BaseField::from(e_vals));
+  }
+
+  result
+}
+
+impl<C: EllipticCurve> ProofOfPossession<C>
+where [(); <C as EllipticCurve>::ScalarField::ORDER]:
```
Autoparallel, line 256: does this `where` clause actually do anything?

src/signatures/bls/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+    let right = pairing::<C, { <C as EllipticCurve>::ScalarField::ORDER }>(pk_ext, pk_ext);
+    if left == right {
+      Ok(())
+    } else {
+      Err(BlsError::VerificationFailed)
+    }
+  }
+}
+impl<C: EllipticCurve> BlsPrivateKey<C>
+where [(); <C as EllipticCurve>::ScalarField::ORDER]:
```
Autoparallel, line 275: Same with this `where` clause

Autoparallel (commented):
Looking a lot better now. We're close. I left a bit more feedback

src/signatures/bls/mod.rs
mahmudsudo: i think this was an error ,this was introduced yesterday to enable extraction of order of the scalarfield , fixed it

src/signatures/bls/mod.rs
mahmudsudo: cleaned up

src/signatures/bls/README.md
mahmudsudo: there is a test for message aggregation and verification

mahmudsudo:
@Autoparallel

mahmudsudo:
@Autoparallel

Autoparallel:
If you could, can you clean up the signatures module now? I think it could be better organized into what the actual algorithms are (i.e., `DSA` isn't really specific and I think the README is a bit redundant).

mahmudsudo:
@Autoparallel

Autoparallel (approved):
Great work!

I talked to @brunny-eth and he will contact you shortly about the bounty.

brunny-eth:
hey @mahmudsudo -- DM-ing you on Telegram per our last discussion

PR #217, opened 2025-02-07
docs: Improve README with more context

voronor:
Hey!

I noticed that the README was a bit brief, so I expanded it with more details on the sum-check protocol, why it's useful, and how to use this implementation. 

Hope this makes it clearer for new users!


Autoparallel:
@voronor I'm trying to be reasonable here, but we just had a pr #206 that moved more of this out into its own section. This basically undoes that.

Issue #30, opened 2024-05-06
FFT and IFFT based on domain.


0xJepsen:

0xJepsen:
@Autoparallel did the Reed Solomon Encoding PR closed this?

kevaundray:
Looking at the PR, I don't think it closes it -- it does a DFT but it is not doing it via the FFT, its doing it via polynomial evaluation

Autoparallel:
Marking as a good first issue as DFT is already done, FFT is not much more than a case for powers of two that can be divide-and-conquered!

mahmudsudo:
can i take on this FFT ?

Autoparallel:
Absolutely

PR #221, opened 2025-02-17
Feat: FFT and Ifft implemented with tests

mahmudsudo:
This PR closes issue #30

### DIFF
--- a/src/polynomial/mod.rs
+++ b/src/polynomial/mod.rs
@@ -258,6 +258,18 @@ impl<F: FiniteField, const D: usize> Polynomial<Monomial, F, D> {
   }
 
   /// Computes the Fast Fourier Transform (FFT) of a polynomial in the Monomial basis.
+  /// 
+  /// of the polynomial in the [`Monomial`] basis by evaluating the polynomial at the roots of
+  /// unity.
+  /// This also converts a polynomial from [`Monomial`] to [`Lagrange`] [`Basis`] with node points
+  /// given by the roots of unity.
+  ///
+  /// ## Returns:
+  /// - A new polynomial in the [`Lagrange`] [`Basis`] that is the result of converting the
+  ///   evaluation of the polynomial at the roots of unity.
+  ///
+  /// ## Panics
+  /// - This function will panic in calling  if the no of coeff isnt a power of two
   pub fn fft(&self) -> Polynomial<Lagrange<F>, F, D>
   where [(); D.is_power_of_two() as usize - 1]: {
     let n = self.num_terms();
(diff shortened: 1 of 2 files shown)


src/polynomial/mod.rs
```diff
@@ -256,6 +256,59 @@ impl<F: FiniteField, const D: usize> Polynomial<Monomial, F, D> {
       }),
     )
   }
+
+  /// Computes the Fast Fourier Transform (FFT) of a polynomial in the Monomial basis.
+  pub fn fft(&self) -> Polynomial<Lagrange<F>, F, D> {
+    let n = self.num_terms();
+    assert!(n.is_power_of_two(), "Length must be power of 2");
```
Autoparallel, line 263: `self.num_terms()` is actually just `D`. 

My suggestion would be to make note of this throughout this whole module and then we don't need assertions that get ran at runtime. We can instead just have compile time checking with const generics. What do you think?

Autoparallel (changes requested):
Do you mind looking into the change I suggested? Could help clean up code elsewhere too. 

I will review more in depth later! Nice work!

src/polynomial/mod.rs
mahmudsudo: i tried some fix , but it might affect some tests

mahmudsudo: @Autoparallel  is the fix in anyway what you want , it will restrict creating odd numbered coefficient polynomials , hence the failing tests

mahmudsudo:
@Autoparallel  any update

src/polynomial/mod.rs
Autoparallel: That only applies to FFT not DFT though

src/polynomial/mod.rs
```diff
@@ -72,6 +72,8 @@ impl<F: FiniteField> Basis for Lagrange<F> {
 }
 
 impl<B: Basis, F: FiniteField, const D: usize> Polynomial<B, F, D> {
+  const CHECK: () = assert!(D.is_power_of_two());
```
Autoparallel: We don't need to have this as an associated constant to the implementation. Instead we can place this only on the FFT.

src/polynomial/mod.rs
```diff
@@ -256,6 +261,58 @@ impl<F: FiniteField, const D: usize> Polynomial<Monomial, F, D> {
       }),
     )
   }
+
+  /// Computes the Fast Fourier Transform (FFT) of a polynomial in the Monomial basis.
+  pub fn fft(&self) -> Polynomial<Lagrange<F>, F, D> {
```
Autoparallel: Here we can have a where clause
```suggestion
  pub fn fft(&self) -> Polynomial<Lagrange<F>, F, D> 
  where [(),D.is_power_of_two() as usize - 1]: 
  {
```

Give this a shot

Autoparallel: This might break compilation if we ever create a polynomial that isn't a power of two though.

Autoparallel:
@mahmudsudo I pushed a commit here for you. We can utilize const generics here to allow compilation of FFT if `D` is a power of two. It's a good trick to know in my opinion. Check it out!

mahmudsudo:
nice trick , didnt cross my mind to convert the bool to usize .

Autoparallel:
Do you think you can add some documentation that will go into the mdBook about the FFT? You can make a directory in the `polynomial` directory for more files other than the `README.md` or something if you need to.

src/polynomial/mod.rs
```diff
@@ -256,6 +256,59 @@ impl<F: FiniteField, const D: usize> Polynomial<Monomial, F, D> {
       }),
     )
   }
+
+  /// Computes the Fast Fourier Transform (FFT) of a polynomial in the Monomial basis.
```
Autoparallel, line 260: Can you make this and the other doc comment like the others in this module? It is lacking.

Autoparallel (changes requested):

mahmudsudo:
@Autoparallel

Autoparallel:
On it.

Autoparallel (approved):

PR #233, opened 2025-04-24
Improve RSA section: link to the examples and style cleanup for file reference

LukeStenger:
This PR updates the RSA section of the book by adding a direct link to the referenced tests.rs file, which contains working examples of RSA key generation, encryption, and decryption.

It also improves the readability of the file reference by applying inline code formatting to tests.rs, along with a slight wording adjustment for increased detail and clarity.

### DIFF
--- a/src/encryption/asymmetric/rsa/README.md
+++ b/src/encryption/asymmetric/rsa/README.md
@@ -21,7 +21,7 @@ Public Key = $(e, n)$
 ## Decryption
 - $m = c^d \mod n$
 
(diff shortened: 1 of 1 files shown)


Autoparallel:
@LukeStenger I appreciate it! 

Usually we want to have contributions be a bit larger. We find lots of bot suggestions today that provide small changes and we want to be sure to differentiate from that.

Do you want to click through the book and see any other spots that could use work? Perhaps it helps you see the repo layout?

